PR #8, opened 2017-02-23
Added Forwarder

craigbr:
Did not add it to docker-compose.yml since it can't be used in a normal deployment

### DIFF
--- a/monasca-forwarder/forwarder_metric_match.yml.j2
+++ b/monasca-forwarder/forwarder_metric_match.yml.j2
@@ -1,0 +1,23 @@
+# The metrics to forward:
+# These can be of three separate forms:
+#    1. Just a name which causes all metrics with that name to be forwarded
+#    2. Just a set of dimensions which causes all metrics with dimensions that
+#       match that set to be forwarded
+#    3. Name and set of dimensions which causes all metrics with that name and
+#       dimensions that match that set to be forwarded
+#
+#    If just a dimension key is given, then the metric matches if it has any
+#    value for that key. If more than one dimension is given, then the metric
+#    must have all of the dimensions.
+#
+# Examples:
+#  - name: storage_usage_mb
+#  - dimensions:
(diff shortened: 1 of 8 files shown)


monasca-forwarder/forwarder_metric_match.yml.j2
```diff
@@ -0,0 +1,23 @@
...
+#  - name: storage_usage_mb
+#  - dimensions:
+#      device: if0
+#      hostname:
+#  - name: network_usage_mb
+#    dimensions:
+#      device: if0
+#
+metrics_to_forward:
+{{ METRICS_TO_FORWARD }}
```
timothyb89: We should probably pass a comma-separated list of metrics in for METRICS_TO_FORWARD and split it, like https://github.com/hpcloud-mon/monasca-docker/blob/master/monasca-agent/templates/prometheus.yaml.j2#L31

As-is the required indentation is a bit funny so it would be good to avoid issues there.

craigbr: The problem is that it is not just metric names that can be configured, it is also metric name + dimensions or just dimensions. While somewhat clumsy, this syntax is better than anything else I could think of using when using dimensions as well.

timothyb89: Hmm, okay, makes sense. Let's merge now and maybe figure something better out later. Thanks!

timothyb89: @craigbr here's an example of metrics and dimensions in a config file: https://github.com/hpcloud-mon/monasca-docker/blob/master/monasca-agent/templates/prometheus.yaml.j2#L12

It accepts syntax like:
```
http://host-a/metrics,http://host-b/metrics|prop=value&prop2=value2,http://host-c
```
We don't need hosts in this case, but using `|` as a separator should work for metric names.

PR #9, opened 2017-02-24
Add k8s file for forwarder

craigbr:
This is in a subdirectory of k8s and not in monasca-start.sh because the
forwarder isn't part of a normal deploy.

### DIFF
--- a/k8s/forwarder/forwarder.yml
+++ b/k8s/forwarder/forwarder.yml
@@ -20,7 +20,7 @@ spec:
             - name: REMOTE_API_URL
               value: http://second_api:18070/v2.0
             - name: METRICS_TO_FORWARD
+            - name: METRICS_TO_FORWARD
               value: "
-  - name: kubernetes.node.allocatable.cpu_agg\n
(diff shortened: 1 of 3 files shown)


k8s/forwarder/forwarder.yml
```diff
@@ -0,0 +1,26 @@
...
+          image: craigbryant/forwarder:latest
+          imagePullPolicy: Always
+          env:
+            - name: MONASCA_PROJECT_ID
+              value: 3564760a3dd44ae9bd6618d442fd758c
+            - name: REMOTE_API_URL
+              value: http://second_api:18070/v2.0
+            - name: METRICS_TO_FORWARD
+              value: "
+  - name: kubernetes.node.allocatable.cpu_agg\n
```
mhoppal: this format is not working for me in kubernetes I changed it to

            - name: METRICS_TO_FORWARD
              value: "
              - name: kubernetes.node.allocatable.cpu_agg\n
              - name: cpu.total_time_sec_agg\n"

and that seemed to work

craigbr:
Does this syntax look right?

PR #10, opened 2017-02-26
Keystone authentication for monasca-api health check

pgillich:
Without Keystone authentication, health check always returns error code. A new script checks it.


monasca-api-python/Dockerfile
```diff
@@ -43,7 +43,6 @@ RUN apk add --no-cache python py2-pip py2-jinja2 curl mysql-client && \
   apk del build-dep
 
 COPY api-* /etc/monasca/
-COPY template.py start.sh  /
```
timothyb89: Ah, good catch. I should probably not be allowed to merge things by hand before I've had coffee :smiley:

timothyb89:
Looks good, thanks for the patch!

monasca-api-python/Dockerfile
pgillich: Never mind, I made similar mistake during my merge...

PR #13, opened 2017-03-06
Use master branch of monasca-transform

handknitted:
Now that kubernetes metrics aggregation is now available and the late metrics
aggregation timing is merged to master we can simplify the docker image
generation to just use the master branch of monasca-transform.


timothyb89:
This looks fine to me, thanks!

As an aside, if you're not opposed, I'd eventually like to port over some of the optimizations from our other Dockerfiles (e.g. layering optimizations, improved git checkouts, configuration from environment, etc) to this component (+spark) as well.

PR #19, opened 2017-03-31
Print travis environment details at the start of CI

timothyb89:


timothyb89:
Still ironing out some details in how travis handles PRs versus branch builds.

`TRAVIS_COMMIT_RANGE` is set correctly for PRs but not branch builds. We don't generally care about branch builds, *except* that we still need them enabled to handle deployment after a PR is merged.

The goals right now are:
 * For `TRAVIS_EVENT_TYPE=pull_request` we should attempt a build based on changed files in `TRAVIS_COMMIT_RANGE`
 * For `TRAVIS_EVENT_TYPE=push` would should attempt to build and push, but only when a PR is being merged. Otherwise, the test should be a no-op pass.
  * The no-op scenario is common for e.g. feature branches. We should ignore them until they are added to a PR.
  * Still need to investigate the exact sets of environment vars needed to tell a feature branch push from a PR merge into master. A follow-up patch will improve on this logic.

timothyb89:
(some PRs like this are going to be forced through while debugging the build script)

PR #32, opened 2017-04-18
Add K8s job cleanup container

timothyb89:
This adds a new job cleanup script/container, intended to clean up old
Kubernetes jobs during an upgrade.

### DIFF
--- a/job-cleanup/Dockerfile
+++ b/job-cleanup/Dockerfile
@@ -1,8 +1,9 @@
 from alpine:3.5
 
-arg CLIENT_TARBALL=https://github.com/kubernetes-incubator/client-python/archive/master.zip
+arg KUBERNETES_CLIENT_VERSION=2.0.0a1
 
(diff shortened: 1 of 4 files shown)


timothyb89:
This job doesn't delete itself quite yet, I will look into adding that now.

job-cleanup/Dockerfile
```diff
@@ -0,0 +1,9 @@
+from alpine:3.5
+
+arg CLIENT_TARBALL=https://github.com/kubernetes-incubator/client-python/archive/master.zip
```
mhoppal: should we be pulling from master or a tagged version so we control code changes to the dependency

timothyb89: 2.0.0a1 (the latest tagged version) is missing some options I'd been trying for cascading deletion. That being said, we can probably switch back to it since cascading delete is totally broken server-side even with library support

mhoppal: lets use the tag then?

PR #36, opened 2017-04-20
Add basic upgrade support to mysql-init

timothyb89:
This add very preliminary 'upgrade' support to mysql init. mysql-init
now automatically creates a schema version table and checks versions
during init. Eventually it will use this table to apply upgrades
sequentially based on the current installed schema version.

Note that actual upgrade logic will come in a future PR. This patch
is mainly intended to make mysql-init return successfully when run
against an existing database. If an existing version is found it will
print the version but will not process the `mysql-upgrades.d`
directory yet.

### DIFF
--- a/mysql-init/mysql-init.d/00-monasca.sql.j2
+++ b/mysql-init/mysql-init.d/00-monasca.sql.j2
@@ -17,8 +17,6 @@
 
 {% set db = MONASCA_DATABASE | default('mon') %}
 
-DROP DATABASE IF EXISTS `{{ db }}`;
-
 CREATE DATABASE IF NOT EXISTS `{{ db }}` DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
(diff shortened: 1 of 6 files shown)


mysql-init/mysql-init.d/00-monasca.sql.j2
```diff
@@ -1,20 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
 */
+
 {% set db = MONASCA_DATABASE | default('mon') %}
 
 DROP DATABASE IF EXISTS `{{ db }}`;
```
mhoppal: we should remove this asap just in case we ever accidentally rerun the script

PR #39, opened 2017-04-24
Add container to create Notifications and Alarm Definitions

craigbr:
Container runs once to create Notifications and Alarm Definitions
in Monasca. The creation command checks first if the Notification
or Alarm Definition exists and matches what is expected. If it
matches, nothing is done so this container is idempotent.

### DIFF
--- a/monasca-alarms/definitions.yml
+++ b/monasca-alarms/definitions.yml
@@ -1,0 +1,193 @@
+#
+# (c) Copyright 2017 Hewlett Packard Enterprise Development LP
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+#
+# Default Notifications and Alarms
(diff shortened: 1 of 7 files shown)


monasca-alarms/definitions.yml
```diff
@@ -0,0 +1,193 @@
...
+    expression: "http_status > 0"
+    match_by:
+      - "service"
+      - "component"
+      - "hostname"
+      - "url"
+    alarm_actions:
+      - root email
+    ok_actions:
+      -  root email
```
kaiyan-sheng: extra space here and the rest of the ok_actions lines.

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -0,0 +1,582 @@
...
+            - Whether the alarm definition should exist.  When absent, removes the alarm definition. The name
+              is used to determine the alarm definition to remove
+    undetermined_actions:
+        required: false
+        description:
+            -  Array of notification names that are invoked for the transition to the UNDETERMINED state
+Examle yaml file:
+
+notifications:
+  - name: Notify via Email
```
timothyb89: :smile:

monasca-alarms/definitions.yml
```diff
@@ -0,0 +1,193 @@
...
+# under the License.
+#
+
+#
+# Default Notifications and Alarms
+#
+notifications:
+  - name: root email
+    type: email
+    address: root@localhost
```
timothyb89: My only real concern here, will there be issues if notifications can't be sent? `root@localhost` will never be valid, so I'd like to be  sure it won't cause trouble if the defaults are used.

craigbr:
The notification engine will try once and if it can't send the email, it will just log a message and drop the notification

PR #43, opened 2017-04-27
Fix error when killing jobs

timothyb89:


job-cleanup/build.yml
```diff
@@ -1,5 +1,6 @@
 repository: monasca/job-cleanup
 variants:
-  - tag: 1.1.0
+  - tag: 1.1.1
     aliases:
+      - :1.1
```
mhoppal: im confused how does this work?

timothyb89: Ah, huh. Seems to be missing something here...

PR #49, opened 2017-05-01
Add dockerfile for monasca-aggregator

rbrndt:


timothyb89:
oh we should get a build.yml going for this

PR #56, opened 2017-05-04
Support forcing a push for an unchanged build.yml via commit message

timothyb89:
Currently `ci.py` won't try to push unless build.yml is changed. This works alright for most standard semver images, but it makes `master-{date}-{time}`-style image tags incompatible with our CI.

This adds support for building using a flag in the commit message, `!push`. When `!push module` is included in a merge commit message (on its own line), `ci.py` will build and push the image regardless of whether or not the `build.yml` has changed.

### DIFF
--- a/ci.py
+++ b/ci.py
@@ -18,9 +18,13 @@
 from __future__ import print_function
 
 import os
+import re
 import signal
 import subprocess
 import sys
+
+
+TAG_REGEX = re.compile(r'^!(\w+)(?:\s+(\w+))?$')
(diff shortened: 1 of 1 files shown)


timothyb89:
Still may need to support more events if we want to build an image without changing build.yml at all. Maybe the travis API?

Hopefully not necessary in most cases, though. If an image depends on an openstack review we'll need to update the XYZ_BRANCH arg to point to the latest revision anyway.

PR #69, opened 2017-05-12
Add print of thresh config file

craigbr:
Also, forces a new build of the thresh container to pick up the
new storm container which has additional configuration env
vars

!push monasca-thresh


timothyb89:
Hmm, looks like the ci script didn't pick up either of the !push-es. I'll push manually and fix that ASAP.

PR #75, opened 2017-05-26
Increase docker-compose version up to 3

kornicameister:
docker-compose in 3rd version offers compatybility with docker's
swarm mode. Since docker-compose in current shape works for
3rd version, the change is painless and allows to deploy monasca
over swarm with Docker (at least) 1.13 as in:

```docker deploy --compose-file docker-compose.yml mon```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '3'
 services:
   influxdb:
     image: influxdb:latest


timothyb89:
Looks good, thanks!

timothyb89:
This may impact our ability to test in Travis since they don't necessarily have the latest version of docker-compose installed. That said, not going to worry about it yet since we don't actually test anything here yet :+1:

PR #77, opened 2017-06-08
Add API access log

craigbr:
It goes to stderr with other logging. It is marked by having
gunicorn.access in the line.

In addition, modify the config file to remove deprecation
warnings.


craigbr:
Need to rebase

PR #78, opened 2017-06-08
Add API access log

craigbr:
It goes to stderr with other logging. It is marked by having
gunicorn.access in the line.

In addition, modify the config file to remove deprecation
warnings.

### DIFF
--- a/monasca-api-python/Dockerfile
+++ b/monasca-api-python/Dockerfile
@@ -10,6 +10,7 @@ ARG REBUILD=1
 ENV CONFIG_TEMPLATE=true \
   LOG_LEVEL_ROOT=WARN \
   LOG_LEVEL_CONSOLE=INFO \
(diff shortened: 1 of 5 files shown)


craigbr:
Creates access log entries on stderr like:
monasca_1               | 2017-06-08 21:50:44 [26] gunicorn.access [INFO] 127.0.0.1 - - [08/Jun/2017:21:50:44 +0000] GET / HTTP/1.1 200 208 "-" "curl/7.52.1" 0.033414

timothyb89:
Seems reasonable enough. Is there a way to turn off the access logs if they're unneeded?

craigbr:
@timothyb89 Now there is

timothyb89:
Seems to work fine for me. My only question is, how much log output should we expect? I'm a bit worried that having access logs turned on by default could nuke servers that don't have log rotation enabled (it isn't by default for k8s / docker).

PR #83, opened 2017-06-14
Create new agent base image

craigbr:
Will be used for separate collector and forwarder images once
this has been merged and the image uploaded to docker hub

This image has the source code, all dependencies and anything
else that would be needed by both the forwarder and collector


monasca-agent-base/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
timothyb89: for what its worth we don't need a build.sh in here... I really need to go through and delete all of these

PR #89, opened 2017-06-15
Docker-compose in ci/cd

mhoppal:


ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
         sys.exit(p.returncode)
 
 
+def update_docker_compose(modules):
+    try:
+        with open("docker-compose.yml") as compose_file:
+            compose_dict = yaml.load(compose_file)
+    except:
+        raise FileReadException('Error reading chart yaml for changed chart')
```
craigbr: This Exception message doesn't seem to match the file being read

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+    if modules:
+        compose_services = compose_dict['services']
+        for module in modules:
+            service_name = MODULE_TO_COMPOSE_SERVICE[module]
+            services_to_update = []
```
craigbr: I believe these 5 lines could be written more simply as:
services_to_update = service_name.split(',')

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+
+    output, err = p.communicate()
+
+    if p.wait() != 0:
+        print('getting current status failed')
+        return False
+    status_output = output.rstrip()
+
+    exit_code, status = status_output.split(":", 1)
+    if exit_code == "0" and status == "exited":
```
craigbr: return exit_code == "0" and status == "exited"

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+    output, err = p.communicate()
+
+    if p.wait() != 0:
+        print('error getting docker id')
+        return ""
+    return output.rstrip()
+
+
+def wait_for_init_jobs():
+    init_status_dict = {"mysql-init": False,
```
craigbr: Why not grafana-init

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+            if docker_id_dict[init_job] == "":
+                docker_id_dict[init_job] = get_docker_id(init_job)
+            if status:
+                amount_succeeded += 1
+            else:
+                updated_status = get_current_init_status(docker_id_dict[init_job])
+                init_status_dict[init_job] = updated_status
+                if updated_status:
+                    amount_succeeded += 1
+        if amount_succeeded == 3:
```
craigbr: Use len(docker_id_dict)

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+            print("All init-jobs passed!")
+            break
+
+    if amount_succeeded != 3:
```
craigbr: See above

ci.py
```diff
@@ -178,12 +206,158 @@ def kill(signal, frame):
...
+        print("Init-jobs did not succeed printing docker ps and logs")
+        output_docker_ps()
+        output_docker_logs()
+        print('Exiting!')
+        sys.exit(1)
+
+    # Sleep incase jobs just succeeded
```
craigbr: s/incase/in case/

ci.py
mhoppal: yep good call

ci.py
mhoppal: i could do that as well as the alarm init. I was thinking of adding a check later on to check that all containers are in a healthy state or exited with 0 so that would catch the other inits and containers

PR #90, opened 2017-06-15
Redirect all grafana logging to stdout

craigbr:
Also bump version


grafana/grafana.ini
```diff
@@ -231,21 +231,21 @@ check_for_updates = true
...
-;mode = console, file
+;mode = console
 
 # Either "trace", "debug", "info", "warn", "error", "critical", default is "info"
 ;level = info
 
 # For "console" mode only
 [log.console]
-;level =
+;level = info
```
mhoppal: should we make this configurable?

timothyb89: All grafana options are configurable via environment variables by default. I believe this could be overridden with `LOG_CONSOLE_LEVEL=warn` set in docker-compose.yml.

PR #92, opened 2017-06-15
Write all kafka logs to stdout

craigbr:
They are discriminated by the filename= attribute on each log line


craigbr:
Do not merge until I check about the filename= attributes

kafka/start.sh
```diff
@@ -83,5 +91,11 @@ fi
 echo "Creating topics..."
 python /create_topics.py &
 
+if [ "$GC_LOG_ENABLED" != "true" ]; then
+  # This turns off the GC logging in /kafka/bin/kafka-server-start.sh
+  # It is a hack, but I could not find another way to do it
+  sed "-i.sv" -e "s/-loggc//" /kafka/bin/kafka-server-start.sh
```
mhoppal: sed for the win!

kafka/templates/log4j.properties.j2
```diff
@@ -15,47 +15,35 @@
 
 kafka.logs.dir=/logs
 
-log4j.rootLogger=INFO, stdout
+log4j.rootLogger={{ SERVER_LOG_LEVEL | default("INFO") }}, kafkaAppender
```
mhoppal: do you have to default here since you did that above in the exports? same as the below?

craigbr: Done. Removed the double defaulting

craigbr:
!push kafka

PR #107, opened 2017-06-27
Use new version of grafana

craigbr:
It needed to be built first so this is a separate change


timothyb89:
Should this update grafana rather than grafana-init?

Also, you should be able to bump the version number in the same commit as your other changes. It will detect the version number change and automatically do a !push for you.

craigbr:
Oops. Should have been grafana, not grafana-init. Fixed

PR #120, opened 2017-07-18
Add memcached for monasca-api

kornicameister:
Memcached is caching tool that allows to limit
the number of requests to the Keystone server
to authorize API access.
It is also the recommended setup for the keystonemiddleware,
used in monasca-api.


kornicameister:
hmm...interesting...looks like memcached runs there somewhere already...port **11211** is already in use :/. Anyway, tell me if you'd be interested in having that, and if so - I will change the port mapping.

kornicameister:
@timothyb89 could you take a look into this one ?

docker-compose.yml
```diff
@@ -1,5 +1,11 @@
 version: '3'
 services:
+
+  memcached:
+    image: memcached:alpine
+    ports:
+      - "22122:11211"
```
kornicameister: Hmm...ok, fair idea. I don't expect we would need an access here. If we you really need I presume it would be simple to just launch some bash container in the same network, have it linked and examine the memcache.

docker-compose.yml
```diff
@@ -90,13 +96,15 @@ services:
     image: monasca/api:master
     environment:
       SIDECAR_URL: http://monasca-sidecar:4888/v1/ingest
+      MEMCACHED_URI: memcached:22122
```
kornicameister: Nice, on it.

PR #123, opened 2017-07-18
Update api port env name

mhoppal:


timothyb89:
FWIW this is a workaround for the (not entirely unlikely) case that the monasca-api container is just named `api`, in which case docker will create its own API_PORT variable overwriting whatever value we configure with some `tcp://...`

PR #127, opened 2017-07-21
Standarize Python components image

kornicameister:


kornicameister:
@timothyb89 let me know what you think of that. I noticed there are couple of common items that are shared between images for python components. Figured an image like this one might help to make images following the same approach everywhere.

It's WIP, but in general I'd be happy to continue this if you're in.

@mhoppal what's your take on that ?

mhoppal:
i think this is awesome!!!!

We have been wanting to do this for awhile now and just have not got around to it.

kornicameister:
@mhoppal I've been playing around this for a while and reached, not too good, conclusion. 
The main problem with base image is that you cannot override those ARG, I've declared in base image, in the way they'd be picked by the parent. Even using **ONBUILD** does not change anything. As **ONBUILD** instructions are executed during **FROM** and **ARG**s are overridden after.

Having that said, I've created that concept where parent only defines entire process:

* dependencies (and removing them)
* adds two script (build.sh & install.sh)

nevertheless in order to have it working, you still need to call the script from the CHILD image. 

Would that be acceptable approach ? I am reluctant to write down anything more if this is not the way you would like to follow.

@mhoppal , @timothyb89  - any suggestions ?

timothyb89:
I don't think calling a script from the child images is a problem, even if it was just adding some utility functions like `clone` it'd be a big improvement to what we have now. Are the `ONBUILD ARG` instructions still necessary with the `install.sh` approach?

One issue I see is that the install process is much more set in stone. In some cases we've needed to make changes before installing dependencies, for instance running `sed` on requirements.txt - if that became necessary in the future it would be difficult to reuse the functions here.

Perhaps `install.sh` could be tweaked slightly to allow it to be imported via `source /install.sh`? `build.sh` could parse arguments and define the normal install+build flow, but if needed we could import the install/clone/cleanup/etc functions and roll a slightly custom build process.

kornicameister:
@timothyb89 I will look into that. Main idea for the requirements I have would be mounting the requirements file and reusing that file instead of the one shipped with the project. Still, note that I (or rather that is requirement Fujitsu has) wants to have those images to follow Openstack constraints. That said, if we would build api,log-api (or any other Python component's image) for, let's say **stable/ocata**, we would use constraints from that Openstack release. That is the main and only purpose of **CONSTRAINTS** variable inside ```install.sh```. Why exactly did you need to change the requirements  ?

For the mutable build process...hmm...I will extract all the functions install.sh defines into own files. That said, child image should be able to override them via ```COPY``` or ```ADD``` instruction.

In overall, after spending some time with this approach, I am starting to see it's limitations and some cumbersomeness. Started to wonder if, for example, creating a generic Dockerfile as extensible Jinja template is not batter idea. Although that would require significant changes in your CI setup, we could have a true template with *Jinja blocks* defined for various part of the process:

* APK dependencies
* constraints control
* install control
* configuration control

etc

For example, in more or less pseudo-code:

```
FROM {{ BASE_IMG | default('alpine') }}:{{ BASE_IMG_TAG | default('3.6') }}

{% block python_version %}
ARG PY_VER=2
{% endblock %}

ARG REPO={{ REPO }}
ARG BRANCH={{ BRANCH | default('master') }}
ARG EXTRA_DEPS= {{ EXTRA_DEPS | default ('') }}
ARG EXTRAS={{ EXTRAS | default ('') }}
ARG CONSTRAINS= {{ CONSTRAINS | default ('https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=BRANCH }}'

{% block apk_deps %}
ARG APK_DEPS={{ APK_DEPS | default('') }}
{% endblock %}

etc...
```

Not really sure, but that is somehow a different approach to the topic. However, as the start I think I am quite happy to roll with the one I initially presented and in future try this approach. Obviously we could build yet-another-image that would enclose, at least, rendering the template.

kornicameister:
Ok, hopefully the image in this shape should be fine enough. I've provided an example to just show the way you could build child image. 

I also decided to use python2/3 alpine images because I had troubles with adding packages via alpine for different python versions. Hopefully you find that addition interesting ;-)

mhoppal:
@timothyb89  can you take a look

monasca-python/build.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
+
+set -x
+
+if [ -e /apk.sh ]; then
+    echo "Overriding default apk dependencies install function"
+    source /apk.sh
+else
+    echo "Usign default apk dependencies install function"
```
kornicameister: It is done ;-)

kornicameister:
@timothyb89 any news ? Did you have a chance to stress that ? ;)

monasca-python/build.yml
```diff
@@ -0,0 +1,10 @@
+repository: monasca/python
+variants:
+  - tag: '2'
+    aliases:
+      - :default
+    args:
+      PYTHON_VERSION: '3'
```
kornicameister: Done

monasca-python/build.yml
```diff
@@ -0,0 +1,10 @@
+repository: monasca/python
+variants:
+  - tag: '2'
+    aliases:
+      - :default
```
kornicameister: Done, also did the same for Python  3

monasca-python/Dockerfile
```diff
@@ -0,0 +1,5 @@
+ARG PYTHON_VERSION="2"
+FROM python:${PYTHON_VERSION}-alpine
```
kornicameister: Done, added also variable that will easily allow to build different variants of the base image for the case where alpine gets bumped. That case we will only need to update ``build.yml`` I guess.

kornicameister: Next PR coming in

monasca-python/build.yml
kornicameister: Next PR coming in

monasca-python/build.yml
kornicameister: Next PR coming in

PR #130, opened 2017-07-24
Allow Thresh to run without storm

craigbr:
Allow the Threshold Engine to optionally run without Storm Nimbus and
Supervisor daemons. Make this the default configuration so remove
Nimbus and Supervisor from docker-compose.yml

Update ci.py for new and removed containers

Fixed CI test for number of jobs


kornicameister:
Works for me.

PR #132, opened 2017-07-27
Kafka image

kornicameister:
chef/kafka seems to follow some Kafka recomendations
for setting up the kafka therefore it only seems fitting
to have and adjust it.

Commits removes custom image building but keeps rest
of the features/tweaks there has been already introduced.


kornicameister:
Obviously **README.md** will have to be updated, but let's collect some initial feedback.

timothyb89:
Obviously there's many factors to consider, but we still have a few reasons to use our image:
 * `ches/kafka:0.9.0.1` is 335 MB compressed vs our 91 MB (and we include a whole python install too!)
   * mostly the savings are from alpine and openjdk-jre-base (safe in testing so far)
   * there are probably licensing issues when using oracle java
     * `ches/kafka` uses [`netflixoss/java:8`](https://hub.docker.com/r/netflixoss/java/) which hasn't been updated in > 2 years and probably violates Oracle's distribution licensing...
 * automatic topic creation (though we want to move this logic elsewhere)
   * we found kafka's built-in automatic topic creation to be very troublesome vs our explicit topic creation script

I think we actually took a lot of ideas from `ches/kafka` when making our kafka image. At the time it didn't support 0.8.x which forced us to make an image from scratch, but I think we ended up with something that better suited our needs. I'm not totally convinced we'd be better off with a community image?

kornicameister:
Ok, a lot to learn for future :). Anyway you know much more about all that, so just sorry if my PRs are worthless :D (well, some has been merged, but in some I see that still is ahead of me). Anyway, I will look into other images and see what's the current state. Might be that I will just add some tweaks to the existing one if I find something worth adding. 

Thx for that feedback. Cheers 👍

kornicameister:
Did not find anything useful, closing down :(

Issue #134, opened 2017-07-28
An approach to kafka-init


kornicameister:
Hi wanted to discuss some of the options about **kafka-init** image we've mentioned a while ago.

Basically, since like ever, the approach I've been taking (which also holds for this repository) to create, or rather manipulate, topics in kafka was to used ```kafka_topics.sh``` shipped together with kafka.
However that means that I'd actually need to install kafka to do this in easiest possible way (might be even to extend an existing image). 

Are you aware of any other tools there might be ?
I will try do some digging but approach lied out above seems, at least now, the only approach possible.

timothyb89:
Huh, I thought our `create_topics.py` created topics through pykafka, but it seems to be just calling out to that same `kafka_topics.sh` script. Guess my memory isn't so good :smile: 

I think you're right though, shelling out to `kafka_topics.sh` seems to be the only real way to do it. Our best bet would be to extend our kafka image with the init tools to at least give Docker/k8s a chance to reuse the downloaded data.

kornicameister:
Yeah, I think I will roll out with an extension of the existing image. Should be fairly straightforward.

kornicameister:
@timothyb89 to be honest, we could have an image to utilize topic creation without extending aforementioned image. We would have to write small Java/Scala App though. See this [question](https://stackoverflow.com/questions/16946778/how-can-we-create-a-topic-in-kafka-from-the-ide-using-api/23360100#23360100).

timothyb89:
Huh, interesting find! I almost want to hack in support for that to pykafka... that would be a huge help in our python bits.

kornicameister:
Hmm...yeah we could do that, but as long as there's no support from API it will take some time :/. 
I haven't seen any, have you ?

kornicameister:
Wrong button :/

timothyb89:
So, to be clear I think the "simple" way forward (for now at least) is to call out to create-topics.sh like everyone recommends, though  That being said..

I read through the code for AdminUtils from the example you linked, it looks like it just makes a bunch of calls out to ZooKeeper. I'd bet it could be replicated in python with kazoo pretty easily... may have to look into that at some point :smile:

kornicameister:
@timothyb89 ok, so for now I will roll with extending the base image and the Python approach we'll leave for later. Once I submit the PR with kafka-init, or prior to, I guess I'll create an issue to keep that future extension in mind.

PR #138, opened 2017-08-01
Remove topic creation from monasca/kafka

kornicameister:
Kafka init

Moved kafka topic creation to external image. That allows reusing the image from different locations.  

Partially_closes #134


kafka-init/README.md
```diff
@@ -0,0 +1,87 @@
...
+    metrics:64:1,alarm-state-transitions:12:1,alarm-notifications:12:1,retry-notifications:3:1,events:12:1,60-seconds-notifications:3:1
+
+If desired, default config parameters can be set using `KAFKA_TOPIC_CONFIG`.
+These use `key=value,key2=value2` syntax and will be translated into
+`--config key=value` arguments as described above. If a duplicate key is passed
+in a specific topic string, it will override the value specified here.
+
+
+[1]: http://semver.org/
+[2]: https://github.com/monasca/monasca-docker/kafka
```
matrixik: This will give 404

kafka-init/README.md
```diff
@@ -0,0 +1,87 @@
...
+Several parameters can be specified using environment variables:
+
+| Variable                      | Default          | Description                           |
+|-------------------------------|------------------|---------------------------------------|
+| `ZOOKEEPER_CONNECTION_STRING` | `zookeeper:2181` | Comma-separated list of ZK hosts      |
+| `KAFKA_HOST`                  | `kafka:9092`     | One of the Kafka brokers              |
+| `KAFKA_TIMEOUT`               | `60`      | How long to wait for Kafka to become available |
+| `KAFKA_CREATE_TOPICS`         | `unset`   | Topics to create on startup, see below       |
+| `KAFKA_TOPIC_CONFIG`          | `unset`   | Default config args for created topics       |
+| `KAFKA_CREATE_TOPICS_SCRIPT`  | `KAFKA_CREATE_TOPICS_SCRIPT` | Path to script that creates topics |
```
matrixik: Should default be `/kafka/bin/kafka-topics.sh`?

kornicameister: Thx, done.

kafka-init/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Partitions and replicas also support an index-based shorthand, so the following
+works as well:
+
+```
+[topic name]:[partitions]:[replicas]
+```
+
+As an example, this is a valid `KAFKA_CREATE_TOPICS` string for [Monasca][8]
+installations as used in the [docker-compose][4] environment:
```
kornicameister: Fixed, also removed unused refs from **kafka/README.md**

kafka-init/README.md
kornicameister: Done, thx

kafka-init/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG KAFKA_VERSION
+ARG SCALA_VERSION
+
+FROM monasca/kafka:${KAFKA_VERSION:-"0.9.0.1"}-${SCALA_VERSION:-"2.11"}
```
timothyb89: I'm not sure if it's a big issue or not, but the pr-bot will not be able to parse this `FROM` tag (and I'm not sure there's a reasonable way to parse it given the the bot's resource model). So this image won't get automatic updates from the base kafka image. We can either live with that (really not too bad) or use a plain `FROM monasca/kafka:0.9.0.1-2.11`.

BTW, any reason to use `${KAFKA_VERSION:-"0.9.0.1"}` rather than just `ARG KAFKA_VERSION=0.9.0.1`? Does docker not evaluate them properly in `FROM` or anything like that?

(it might also be good to depend on the explicit container version somehow, e.g. `monasca/kafka:${KAFKA_VERSION}-${SCALA_VERSION}-1.0.2`)

kafka-init/build.yml
```diff
@@ -0,0 +1,8 @@
+repository: monasca/kafka-init
+variants:
+  - tag: 0.0.1
+    aliases:
+        - ':latest'
```
timothyb89: super nit pick, this line is over-indented

timothyb89:
This seems to work fine in docker-compose. We'll need to bump the version in `kafka/build.yml` to push a new image, though. I'd say it should at least be a minor version bump (`1.0.2` -> `1.1.0`). I made some nodes about versioning concerns but I'm not sure if it's worth changing, I'll let you decide :smile: 

Also, we'll need to coordinate things a bit with monasca-helm - can't bump the kafka version there until we have a kafka-init chart!

kafka-init/build.yml
kornicameister: Done

kafka-init/Dockerfile
kornicameister: Replaced that with single ARG. Basically there's no point in having multiple variables as we inherit from the ```monasca/kafka``` image here. Should be much simpler now and easier to parse.

kornicameister:
Upgraded minor version to ```1.1.0```.
About the Helm...sadly the requirements has changed in a way where we don't want to go with Helm for now. If you really want me to handle that bit, I will do that - it's not in anyone's intention to mess with someones backyard.  Just let me know what you think about that..

kornicameister:
;-(...seems like it does not build because it cannot find locally build (or kafka-init build happens prior to tagging new monasca/kafka). ```monasca/kafka:0.9.0.1-2.11-1.1.0```.

Are there any changes required in the ci.py or dbuild to pick up such nits and organize the build or should I split the PR into 2 ?

artur-ba:
Tested this in my local env and it works. Topics were created in Kafka and I was able to list them.

kafka/build.yml
```diff
@@ -1,7 +1,12 @@
 repository: monasca/kafka
 variants:
+  - tag: 0.9.0.1-2.11-1.1.0
+    aliases:
+      - :latest
+    args:
+      KAFKA_VERSION: "0.9.0.1"
+      SCALA_VERSION: "2.11"
   - tag: 0.9.0.1-2.11-1.0.2
```
timothyb89: This tag should actually be removed or it will overwrite the existing image (needed in helm for now) with the same content as the new version.

timothyb89:
re: helm, that's no problem, we'll get around to adding it eventually - this has been on our to-do list for a while now, so you've already saved us a good bit of work :smile:

As for the failure, it looks like you're right, kafka-init is being built before the new kafka image. Unfortunately the only way to guarantee they'll build in the correct order is to split the PR.

It would be nice to support some dependency chains but that would require changes to both the CI script and dbuild, since dbuild doesn't guarantee build order (as it may build in parallel). I filed timothyb89/dbuild#2 to track improvements to this, but I likely won't be able to prioritize it for a while.

kornicameister:
Ok, I've splitted the PR, but unfortunately ```0.9.0.1-2.11-1.0.2``` is being rebuilt in CI and the ```Dockerfile``` meant for ```0.9.0.1-2.11-1.1.0``` is also used for ```0.9.0.1-2.11.1.0.2``` hence no topic are actually created inside the CI run. 

I expected only ```0.9.0.1-2.11-1.1.0``` to be build. Is this expected behaviour ?

timothyb89:
The variant for `...-1.0.2` should be removed entirely from build.yml, it will always try to (re)build (and push) all variants any time build.yml changes (or `!push` is included in a merge commit).

kafka/build.yml
```diff
@@ -1,5 +1,9 @@
 repository: monasca/kafka
 variants:
+  - tag: 0.9.0.1-2.11-1.1.0
+    args:
+      KAFKA_VERSION: "0.9.0.1"
+      SCALA_VERSION: "2.11"
   - tag: 0.9.0.1-2.11-1.0.2
```
timothyb89: This `-1.0.2` variant should be removed from build.yml entirely

timothyb89:
As an FYI to any interested parties (@mhoppal) , we'll probably need to force this PR through since CI should fail. #153 is required to get topic creation working again, so merging this will break master until kafka-init is in.

kafka/build.yml
kornicameister: Ok, I've removed the variant.

kornicameister: @timothyb89 is it like an "official approach"- we should drop older entries in ```build.yml``` to disallow overriding older images with changes meant for newer ones ?

timothyb89: @kornicameister I think so, yes. It's less of an issue for images that use date+time stamped tags since they'll always end up unique, but we definitely want to avoid introducing breaking changes by overwriting a tag that looks like it should be unique (like semver)

timothyb89:
FYI master will be broken until #153 merges.

PR #141, opened 2017-08-04
Add keystone-init container

timothyb89:
This adds a new container capable of creating accounts in keystone.
It randomly generates passwords at first start and stores them in
Kubernetes secrets, bypassing Helm so that the secrets can be stored
securely.

Note that the container currently requires a Kubernetes environment,
support for running in docker will be added in the future. Eventually
this should replace most of our keystone bootstrapping logic.

### DIFF
--- a/keystone-init/Dockerfile
+++ b/keystone-init/Dockerfile
@@ -11,6 +11,6 @@ RUN apk add --no-cache ca-certificates && \
     rm -rf /root/.cache/pip && \
     apk del build-dep
 
-COPY keystone_init.py kubernetes.py preload.yml /
+COPY keystone_init.py kubernetes.py preload.yml start.sh /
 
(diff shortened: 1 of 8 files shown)


timothyb89:
@kornicameister short term we're planning to use this only in Kubernetes/Helm. Kubernetes has its own (reasonably robust) secret management that we aren't using properly right now, since all our passwords get stored plaintext in Helm's config. Ideally we'd like to use this in Docker too, but it will be limited to just creating accounts since Docker has no real secret management and no API to manage it inside the cluster.

This PR should be enough to improve our Kubernetes deployment, but some work is still needed to run in Docker:
 - the hard dependency on the Kubernetes API should be removed
 - `keystone_init.py` needs to support creating keystone services and endpoints
 - bootstrapping logic should be removed from the `keystone` container

PR #142, opened 2017-08-07
Couple fixes for monasca/python

kornicameister:
Short flags, composed out of two letters, do not work
with getopt utility.


kornicameister:
@timothyb89 I think we need to pull it first, otherwise I won't be able to specify correct tags in other changes. Thx ;)

PR #144, opened 2017-08-07
Changed monasca-grafana app repo link

mohamhab:


grafana/build.yml
```diff
@@ -3,3 +3,8 @@ variants:
   - tag: latest
     aliases:
       - :4.0.0-1.1.1
+  - tag: app
+    aliases:
+      - :4.0.0-1.2.0
+    args:
+         MONASCA_APP_REPO: https://github.com/monasca/monasca-grafana.git
```
timothyb89: Should be a 2 space indent under `args`, e.g.
```
args:
  MONASCA_APP_REPO: ...
```

grafana/build.yml
```diff
@@ -3,3 +3,8 @@ variants:
   - tag: latest
     aliases:
       - :4.0.0-1.1.1
+  - tag: app
+    aliases:
+      - :4.0.0-1.2.0
```
timothyb89: Actually, thinking about this more, we should put an extra suffix on the alias, probably` - 4.0.0-1.2.0-app`. We can remove the suffix once we move to the branch permanently.

PR #148, opened 2017-08-09
Rebuild monasca-python

kornicameister:
!push monasca-python


kornicameister:
@timothyb89 no idea if that's the correct way to enforce availability of new tags in the hub for ```monasca/python```, but if it is - yeah :dancing_men:

timothyb89:
So currently it only checks the merge commit for the `!push` tag (else we could end up releasing accidentally due to a !tag somewhere in a huge chain of commits). Eventually I'd like to rework this to use GitHub labels (same permissions, much better visibility) and probably with some bot integration as well. Can push this through in the mean time :smile:

PR #155, opened 2017-08-11
Tempest in CI

kornicameister:
Add execution of the tempest inside the CI setup.


timothyb89:
We'd originally planned to just use the smoke tests since to keep runtime low, but it looks to be pretty fast anyway (~15 minutes). Nice improvement!

PR #161, opened 2017-08-16
Removed build.sh scripts

### LINKED ISSUE #124: Remove old build scripts
The old `build.sh` scripts are deprecated now in favor of dbuild with `build.yml`.

Related to #84 since we really need docs for dbuild and our CI/CD pipeline in general.

kornicameister:
build.sh are no longer used. Local builds
can be executed with plain `docker build`
command therefore those scripts are not required.

Closes #124


kornicameister:
@timothyb89 quess the reason it terminated....is....that the log output was too long :/. Anyway, I don't see this as the blocker, those scripts are, as far as I know, are pretty much redundant.

timothyb89:
Yikes, I guess dbuild console output can get pretty spammy when lots of images are built at once. Those 
build scripts are definitely not used in CI. Rebuilding the containers isn't even necessary in this case, I believe we can just skip travis for this commit as a workaround with `[skip ci]`.

Also filed timothyb89/dbuild#3

monasca-python/build.sh
```diff
@@ -1,122 +0,0 @@
-#!/bin/sh
```
timothyb89: Oh wait a sec. Maybe this shouldn't be removed? :smiley:

kornicameister: LOL, :1st_place_medal: for me.
I ran a ```find``` command and just removed everything that popped out...ah :D

kornicameister:
Ok, I had to ``--force``` push, but I guess it is fine now.

PR #143, opened 2017-08-07
One Python to rule them all

kornicameister:
Rewrite all the images to leverage newly added [monasca/python](https://hub.docker.com/r/monasca/python/) image.

PR #162, opened 2017-08-16
[WIP] Provide stack-fix & tini init

### LINKED ISSUE #159: Use tini init
We should use [tini init](https://github.com/krallin/tini) so we can handle signals correctly. ctrl+c on many of our python images gets eaten since we don't have a real init and it just gets absorbed by sh/bash.

Example: [keystone-init Dockerfile](https://github.com/monasca/monasca-docker/blob/master/keystone-init/Dockerfile#L16)

### LINKED ISSUE #160: Apply stack fix to all python containers
musl's default stack size is very small (80 KiB) which can cause frequent stack overflows (probably why the agent-collector crashes). See also: https://github.com/voidlinux/void-packages/issues/4147

A workaround was proposed in https://github.com/esnme/ultrajson/issues/254#issuecomment-314862445. I've had good results when testing it in the [keystone-init container](https://github.com/monasca/monasca-docker/blob/master/keystone-init/Dockerfile#L8) by building the `.so` as part of the container build process, and we should apply it elsewhere - at least to the agent, which likely needs it most.

Basic steps to implement:
 - add `stack-fix.c` to container
 - build with `gcc -shared -fPIC /stack-fix.c -o /stack-fix.so`
 - run python with `LD_PRELOAD=/stack-fix.so python ...`

kornicameister:
Commits fixes two issues:

* #160 - stack-fix.c
* #159 - tini-init

Closes #160
Closes #159
Depends-On #143 

!push monasca-python


kornicameister:
Ah, it won't be that easy to built-it in. I just checked locally and the error was:

```
ERROR: for monasca  Cannot start service monasca: oci runtime error: container_linux.go:247: starting container process caused "exec: \"LD_PRELOAD=/stack-fix.so\": stat LD_PRELOAD=/stack-fix.so: no such file or directory"
```

having ```ENTRYPOINT ["LD_PRELOAD=/stack-fix.so", "/sbin/tini", "--"]``` in **monasca/python** and 
```CMD ["/start.sh"]``` in **monasca/api**

File: monasca-python/Dockerfile [hunk 1]
```diff
@@ -3,5 +3,12 @@ ARG ALPINE_VERSION="3.6"
...
-COPY build.sh /
+COPY build.sh stack-fix.c /
 RUN chmod +x /build.sh
+
+RUN apk add --no-cache ca-certificates tini && \
+  apk add --no-cache --virtual build-dep musl-dev linux-headers git make g++ linux-headers && \
+  gcc -shared -fPIC /stack-fix.c -o /stack-fix.so && \
+  apk del build-dep
+
+ENTRYPOINT ["LD_PRELOAD=/stack-fix.so", "/sbin/tini", "--"]
```
timothyb89: I think we'd need to be running in a shell for `VAR=value` to work (which `ENTRYPOINT` is not). So we probably need:

```
ENTRYPOINT ["/sbin/tini", "--"]
CMD ["LD_PRELOAD=/stack-fix.so", "python", "foo.py"]
```

The LD_PRELOAD override will probably need to be done inside each container since it needs to be in the `CMD` directive or some startup script. We could use `ENV LD_PRELOAD=/stack-fix.so` but I'd almost prefer that not to be user-overridable...

kornicameister: Couldn't we just ```entrypoint.sh``` like so:

```sh
#!/bin/sh
export LD_PRELOAD=/stack-fix.so
/sbin/tini -- "$@"
```

that's raw code, have no idea if it will work but if it does, will it be acceptable approach ?

timothyb89: Sure, I think that would work fine.

On a related note, after reading https://github.com/krallin/tini/issues/8 a bit closer I wonder if we'll end up un-fixing tini's signal handling with our startup scripts. Will need to do some testing to verify.

kornicameister:
Ok, I've added the entrypoint.
To see it in action, you can check out

```sh
cd monasca-python
docker build -t monasca/python:2-test --build-arg PYTHON_VERSION=2 .
docker start -a -i $(docker create -t -i monasca/python:2-test python -c 'import os ; print(os.environ.get("LD_PRELOAD"))')
```

it will print out the path to the compiled c file.

PR #165, opened 2017-08-18
Add default keystone _member_ role

timothyb89:


timothyb89:
hold up this is still broken

Issue #164, opened 2017-08-17
Create Monasca credentials and endpoint in OpenStack's Keystone


matrixik:
Hi,

I'm looking now into using separate Keystone from client OpenStack instance.
Connection to this instance need to be configurable.

I'm looked into `keystone-init` container (thx for creating it) and saw you started adding options to add users, project etc. In README you wrote `The default config file is empty, and should be overridden for this container to be of any use.`
Now in the `keystone` container you have `preload.yml` file with default users and endpoints.
Also `keystone-bootsrap.sh` that connect to Keystone and call script to load everything into Keystone.

Now some questions:
What would be way forward with preloading all data into Keystone?
Should this `preload.yml` file (and `preload.py`) be moved to `keystone-init` container or all data from `preload.yml` should now reside in `docker-container.yml`?

`keystone-bootsrap.sh` should probably also be moved to `keystone-init` container.

Also I'm not sure how to configure access to external Keystone in `docker-container.yml` so that other containers still would be able to use: `depends_on: - keystone`.
When I try removing `image:` from `keystone:` then Docker will complain about that it needs `image` or `build`. As of now I'm still very new to Docker and I'm not sure what to use here. Maybe some script would be needed to check if connection to this external Keystone success?
When I search some Google there are also some people suggesting using [compose `networking`](https://docs.docker.com/compose/networking/) component for this, but it got me really confused if it's needed in this case...

@timothyb89

timothyb89:
You can use a volume mount in docker-compose.yml to provide preload.yml with keystone users without needing to bake it into the container directly. We did this until recently: https://github.com/monasca/monasca-docker/blob/792576e88de418deb0f5e09200f9c41e7ea5dd30/docker-compose.yml#L74-L90
I changed the keystone container to bake in the preload.yml a little while ago, but to be honest I don't think that was a good change, and we'll probably go back to using an external preload.yml for keystone-init.

(I'll note that the new keystone-init script/container doesn't support adding endpoints and services quite yet ... I think we'll need that before we can switch over completely)

The keystone bootstrapping process might need to stay inside the keystone container since it (I think) needs to modify some config files and the database directly, though maybe there are ways around this.

As for configuring network dependencies... docker-compose's dependencies are mostly not useful, to be honest. You can define the container starting order (really not meaningful) but it doesn't actually wait for the service to become "available" - whatever "available" happens to mean for that particular service. For example, monasca-agent should only consider keystone to be "available" if:
[...]

PR #170, opened 2017-08-22
Check Keystone status when starting keystone-init

matrixik:
Before initialising Keystone make sure it's working.


matrixik:
First small change to make sure I'm understanding how all of this works.

matrixik:
Related: #164

kornicameister:
Isn't like this issue will resolve_ #164 ?

matrixik:
No, keystone-init still doesn't support adding endpoints and services.

kornicameister:
don't you think you should call this new method :D

matrixik:
Fixed.

timothyb89:
For what it's worth keystone-init will already retry the initial connection to keystone due to the `@retry()` annotation, though the number of attempts/delay probably should be increased for `get_keystone_client` because we expect it to fail more often: https://github.com/monasca/monasca-docker/blob/e89c2c789a5e280186ca117d1668e057c03c51da/keystone-init/keystone_init.py#L134-L137

That said, we can add an additional check if you think it would help.

matrixik:
Hmmm, you are right. I will just increase numbers for `get_keystone_client`.

matrixik:
Updated with same numbers like in other checks.

PR #172, opened 2017-08-22
Add info about using external Keystone

matrixik:
Example show connecting to Keystone that use aliases for communication
instead of using ports.

For now I tested it with Keystone from freshly installed Devstack. I did not run `keystone-init` on it.


keystone-init/README.md
```diff
@@ -174,6 +174,63 @@ containers:
 {{ include "keystone_env" .Values.my_pod.auth | indent(6) }}
 ```
 
+Using external OpenStack Keystone
```
kornicameister: I think that this documentation, presented in this form, is better to be put to ```monasca-docker/README.md``` instead of here. We have a bunch of examples how to connect to keystone and each of them is not about ```keystone-init```

README.md
```diff
@@ -108,11 +108,68 @@ export MON_AGENT_COLLECTOR_VERSION=ocata
...
+instead of ports when overwriting environment variables:
+
+#### `docker-compose.yml`
+
+In `agent-forwarder` and `agent-collector`:
+```yaml
+environment:
+  OS_AUTH_URL: http://192.168.10.6/identity/v3
+  OS_USERNAME: monasca-agent
+  OS_PASSWORD: password
```
knakazono: I think it's better to add "OS_PROJECT_NAME" here.

matrixik: Adding in https://github.com/monasca/monasca-docker/pull/179

PR #176, opened 2017-08-25
Sync containers time with host time

matrixik:
This will link host machine time to time in the containers so we will have same time in docker output logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,18 @@ environment:
   KEYSTONE_ADMIN_PASSWORD: secretadmin
 ```
 
+Synchronizing containers time with host time
+--------------------------------------------
+
+If you want to synchronize containers time with host time they are running on,
+you need to add to every service following volumes:
+
+```yaml
+volumes:
+  - "/etc/timezone:/etc/timezone:ro"
+  - "/etc/localtime:/etc/localtime:ro"
+```
+
 [1]: https://docs.docker.com/compose/
 [2]: https://github.com/hpcloud-mon/monasca-docker/blob/master/keystone/preload.yml
 [3]: https://docs.docker.com/compose/environment-variables/


matrixik:
Done

kornicameister:
@timothyb89 works ok for me, all the timestamp everywhere feature the host timestamp.
if you feel like this is nice addition, please merge :)

timothyb89:
Hmm, I think we'd prefer to keep timestamps as UTC by default. Maybe we could make this optional? e.g. an override .yaml (`docker-compose -f docker-compose.yml -f local-time.yml`) or document the change somewhere?

kornicameister:
I think I'd opt for the documentation, simply because we would need to have overrides for each service and it would make it a bit cumbersome to apprehend. Documentation, however, that would show how we can achieve that with appropriate extension file sounds much more user friendly.

matrixik:
Looks like to provide this separate file I would need to create to separate files: `local-time.yml` and `local-time-log.yml` because I can't add both of them into one file. With one file `docker-compose -f docker-compose.yml -f local-time.yml` will fail because:
```
ERROR: The Compose file is invalid because:
Service kibana has neither an image nor a build context specified. At least one must be provided.
```
I'm not sure if I want to add two separate files for such task.

I can see 3 ways forward:
1. I could add two files if you think is fine
2. I could add only one file with log pipeline commented out
3. I can only add info to documentation about this

@timothyb89 Thoughts?

kornicameister:
With all those problems, I'd opt for documentation solution.
In overall, I believe that docker-compose files here are only relevant for this repository and CI deployment, for all other cases we should craft the compose file on our own to have it suits the needs the best way.

+1 for docs

timothyb89:
Docs seems reasonable to me as well. Thanks!

matrixik:
Changes to `yaml` files removed, added info to README.

PR #177, opened 2017-08-25
Enable kibana.yml templating

kornicameister:
The configuration file for kibana was not templated at all
during the startup of the container. Thus the environmental
variables were useless.

This commit adds python and python2-jinja packages to
enable templating similar to other monasca images.


kibana/Dockerfile
```diff
@@ -31,13 +31,17 @@ WORKDIR /
...
+COPY wait-for.sh start.sh template.py kibana.yml.j2 /
+
+ARG REBUILD_DEPENDENCIES=1
+RUN apt-get update -qq && \
+  apt-get install python python-jinja2 -y -qq && \
+  chmod +x /wait-for.sh /start.sh /template.py
 
 ARG REBUILD_INSTALL_PLUGIN=1
-RUN /plugin_config && \
+RUN python /template.py /kibana.yml.j2 /opt/kibana/config/kibana.yml && \
```
timothyb89: Does it make sense to run the template script here? Normally we only run it at runtime and not during build.

kornicameister: the annoying part of the kibana and plugin installation is that they require presence of its configuration therefore I had to run this once prior to plugin installation.

Later on it is executed same as for every other container/image,

PR #185, opened 2017-08-28
Add keystone-shell container

timothyb89:
The keystone-shell is a container to help debug keystone instances in
Kubernetes. It provides a ptipython-based shell with highlighting,
autocomplete, and access to the Keystone and Kubernetes APIs, along
with an OpenStack client and utility function to read and swap between
credentials from secrets (in any namespace) at runtime.

### DIFF
--- a/keystone-shell/Dockerfile
+++ b/keystone-shell/Dockerfile
@@ -1,0 +1,23 @@
+FROM python:3.6-alpine3.6
+
+# To force a rebuild, pass --build-arg REBUILD="$(DATE)" when running
+# `docker build`
+ARG REBUILD=1
+
+COPY requirements.txt stack-fix.c /
+
+RUN apk add --no-cache ca-certificates tini libffi && \
+    apk add --no-cache --virtual build-dep \
+        musl-dev linux-headers git make g++ libffi-dev openssl-dev && \
+    gcc -shared -fPIC /stack-fix.c -o /stack-fix.so && \
+    pip install urllib3 ipaddress ipython ptpython python-openstackclient && \
+    pip install -r /requirements.txt && \
+    rm -rf /root/.cache/pip && \
+    apk del build-dep
+
+COPY keystone_shell_vars.py kubernetes.py ptpython_init.py start.sh /
+COPY keystonerc.sh /root/.keystonerc
+
+ENV ENV="/root/.keystonerc"
+ENTRYPOINT ["sh", "-l", "/start.sh"]
+
--- a/keystone-shell/README.md
+++ b/keystone-shell/README.md
@@ -1,0 +1,106 @@
+keystone-shell
+==============
+
+This image allows you to quickly interact with a keystone environment.
+
+Sources: [keystone-shell][1] &middot; [Dockerfile][2] &middot; [monasca-docker][3]
+
+Tags
+----
+
+Images in this repository are tagged as follows:
+
+ * `latest`: refers to the latest stable point release, e.g. `1.0.0`
+ * `1.0.0`, `1.0`, `1`: standard semver tags
+
+Basic Usage
+-----------
+
+This is mainly intended for use in Kubernetes environments configured using
+[keystone-init][4]. To use, run:
+
+    kubectl run keystone-shell -i -t \
+        --rm=true \
+        --restart Never \
+        --image=monasca/keystone-shell:latest
+
+(pass `-n <NAMESPACE>` as needed if the desired secrets are located elsewhere)
+
+Once the container starts, press enter once to start the shell. Basic usage
+information will be printed to the terminal.
+
+There are three commands available:
+ - `secret <NAME>`: load keystone `OS_` variables from secret in current 
+   namespace `NAME`
+ - `shell [NAME]`: start a Python shell (using ptipython, with highlighting 
+   and autocomplete) with pre-connected Keystone and Kubernetes clients -
+   additional usage information will be printed on startup
+   - if specified, secret with `NAME` will be loaded first (as with 
+     `secret NAME`) and will override existing environment variables
+ - `openstack ...`: the standard openstack client 
+
+Direct Shell
+------------
+
+You can also start a shell directly:
+
+    kubectl run keystone-shell \
+        --rm=true \
+        --restart Never \
+        --image=monasca/keystone-shell:latest \
+        --env="OS_USERNAME=admin" \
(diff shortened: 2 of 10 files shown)


keystone-shell/README.md
```diff
@@ -0,0 +1,106 @@
...
+zero. If Keystone is not available or the credentials are invalid, the return
+code will be 1. Log information should be printed to help narrow down the
+error, or at least as much as Keystone will be willing to divulge about it. 
+
+Configuration
+-------------
+
+| Variable           | Default          | Description                     |
+|--------------------|------------------|---------------------------------|
+| `LOG_LEVEL`        | `INFO` | MySQL server host               |
```
matrixik: Wrong descriptions in many lines.

keystone-shell/keystone_shell_vars.py
```diff
@@ -0,0 +1,152 @@
...
+    if namespace is None:
+        namespace = get_current_namespace()
+
+    try:
+        return client.get('/api/v1/namespaces/{}/secrets/{}', namespace, name)
+    except HTTPError as e:
+        if e.response.status_code != 404:
+            raise
+
+        return None
```
matrixik: I think this return should have one less indentation level. I know you run it from `except` but this way you will make it default.

keystone-shell/keystone_shell_vars.py
```diff
@@ -0,0 +1,152 @@
...
+        secret_name = sys.argv[1]
+
+        if len(sys.argv) > 2:
+            secret_namespace = sys.argv[2]
+        else:
+            secret_namespace = None
+
+        # purge existing keystone vars from the environment
+        for var in keystone_env_vars().keys():
+            print('unset {};'.format(var))
```
matrixik: Should have some printed info that user need to use it.

keystone-shell/keystone_shell_vars.py
```diff
@@ -0,0 +1,152 @@
...
+
+        logger.info('now using account from secret %s', secret_name)
+    else:
+        logger.info('no secret name arg provided, will use default environment')
+
+
+if __name__ == '__main__':
+    main()
+
+
```
matrixik: Too many empty lines at the end of file.

keystone-shell/keystone_shell_vars.py
```diff
@@ -0,0 +1,152 @@
...
+            print('unset {};'.format(var))
+
+        secret = get_kubernetes_secret(secret_name, secret_namespace)
+        for key, val in secret.data.items():
+            val_bytes = base64.b64decode(val)
+            print('export {}="{}"'.format(key, val_bytes.decode('utf-8')))
+
+        logger.info('now using account from secret %s', secret_name)
+    else:
+        logger.info('no secret name arg provided, will use default environment')
```
matrixik: Maybe you could move it up with reversed `if len(sys.argv) > 1:` to have it close to the place why did it run and not with just bare `else` after a lot of code.

keystone-shell/keystone_shell_vars.py
matrixik: Oh, you run this file output with `eval`.

keystone-shell/README.md
timothyb89: wow, looks like my copy&pasting skills are getting rusty!

keystone-shell/keystone_shell_vars.py
timothyb89: Not that it matters a ton since the code paths are effectively equivalent but my feeling was that (without the conditional to check for a 404) is a bit more clear about the intent.

Ideally it would've been:
```
try:
    return client.get(...)
except NotFound:
    return None
````

... which I think makes the intent more clear - that is, `None` isn't the "default" branch so much as the "not found" branch. Since requests doesn't have exception subclasses for different error codes, it gets a bit uglier when we compare the error code ... I'm not sure how much more readable the end result is, though.

File: keystone-shell/keystone_shell_vars.py (hunk 4 above)
Comment (timothyb89) [reply keystone-shell/keystone_shell_vars.py]: good catch, fixed

File: keystone-shell/keystone_shell_vars.py (hunk 5 above)
Comment (timothyb89) [reply keystone-shell/keystone_shell_vars.py]: I think you're right, it's more clear to write that a bit differently.

File: keystone-shell/Dockerfile [hunk 6]
```diff
@@ -0,0 +1,23 @@
+FROM python:3.6-alpine3.6
```
kornicameister: Can't we just use the monasca/python.
It should have all musl and stack-fix included already AFAIR.

keystone-shell/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# `docker build`
+ARG REBUILD=1
+
+COPY requirements.txt stack-fix.c /
+
+RUN apk add --no-cache ca-certificates tini libffi && \
+    apk add --no-cache --virtual build-dep \
+        musl-dev linux-headers git make g++ libffi-dev openssl-dev && \
+    gcc -shared -fPIC /stack-fix.c -o /stack-fix.so && \
+    pip install urllib3 ipaddress ipython ptpython python-openstackclient && \
```
kornicameister: With monasca/python we could always install latest master, but I am not really sure if that's needed at all ?

keystone-shell/keystone_shell_vars.py
timothyb89: yup, the user won't ever run this script directly, the bash alias will take care of evaling the output for them.

keystone-shell/Dockerfile
timothyb89: Yeah, you're right, this should really be using monasca/python. The docker stuff is mostly a copy&paste hack job that came out of my frustration while debugging keystone-init.

I'm fine with holding off on merging until it's converted, though.

kornicameister:
I have really bad and nasty problems trying to make it run.
Most of them resolves around the keystone secret (tried to used some of those that **monaca-helm** has used) with ```kubectl get secret -n monitoring``` but without a success. Finally managed to launch that without ```KEYSTONE_SHELL``` and everything was fine until I tried to get a token from connected keystone. The error encountered was:

```sh
[Errno 2] No such file or directory: '/usr/local/lib/python3.6/site-packages/ptyprocess-0.5.2.dist-info/top_level.txt'
```

From within the container I called ```openstack token issue````

matrixik:
Monasca/python changes are merged so this change could be updated to reflect that.

PR #189, opened 2017-08-31
Add travis badge

matrixik:


matrixik:
Also what do you think about adding some static code analysers (flake8, pylint, ShellCheck) to travis builds?
Or https://houndci.com/ ?

timothyb89:
I think it's a good idea. I haven't looked at Hound too much but it could be a nice option. That being said, it doesn't support shellcheck which is a big chunk of the code in this repo...

PR #192, opened 2017-09-01
Trigger new build of monasca-agent-base container

craigbr:
Need the prometheus fix from https://review.openstack.org/499835

!push monasca-agent-base


craigbr:
Didn't do this right

PR #196, opened 2017-09-06
Create services and endpoints with keystone-init

matrixik:
First shot on https://github.com/monasca/monasca-docker/issues/164

Still not loading all endpoint interfaces, I need to look more into it.
Also docs need to be updated.

!push keystone-init


matrixik:
@timothyb89 please look into code and tell me if I'm going in the right direction.

keystone-init/keystone_init.py
```diff
@@ -316,6 +316,74 @@ def get_or_create_group(client, domain, name):
...
+    """Get service or create it if doesn't exist.
+
+    :type client: keystoneclient.v3.client.Client
+    :type name: str
+    :type service_type: str
+    :type description: str
+    :return:
+    :rtype: keystoneclient.v3.services.Service
+    """
+    services = client.services.list()
```
timothyb89: I think we should cache this response like we do in the other methods.

keystone-init/keystone_init.py
```diff
@@ -316,6 +316,74 @@ def get_or_create_group(client, domain, name):
...
+    endpoints = client.endpoints.list()
+
+    for endpoint in endpoints:
+        logger.debug('existing endpoint %r', endpoint)
+        if endpoint.service_id == service.id:
+            if endpoint.url == url:
+                if endpoint.interface == interface['name']:
+                    return endpoint
+            else:
+                client.endpoints.delete(endpoint)
```
timothyb89: I'm not sure deleting endpoints that already exist is desirable, but if that's really the behavior we want we should at least log a warning about it.

Generally I'd prefer to create or update resources when the preload input specifies but leave everything else alone, unless there's some cases where an extra endpoint could cause issues. So far we've been able to get away without deleting anything, which is a lot less scary IMO.

keystone-init/keystone_init.py
```diff
@@ -316,6 +316,74 @@ def get_or_create_group(client, domain, name):
...
+
+    :type client: keystoneclient.v3.client.Client
+    :type service: keystoneclient.v3.services.Service
+    :type url: str
+    :type interface: dict[str, str]
+    :type region: str
+    :return:
+    :rtype: keystoneclient.v3.endpoints.Endpoint
+    """
+    endpoints = client.endpoints.list()
```
timothyb89: we can cache this too :smiley:

keystone-init/keystone_init.py
matrixik: Changed to only updating existing endpoints

matrixik:
Shitty github review ate my comment...

I have problem with using cache for creating/updating endpoints. When I'm using `_endpoint_cache` it's duplicating endpoints when I change `url` in config between runs on same Keystone, when I retrieve endpoints every time from Keystone it's working properly.
https://github.com/monasca/monasca-docker/pull/196/files#diff-c3bc2b30c557a9e894282ce000d23f10R368

I wasted way too much time today to make it work and it's possible that I'm doing some stupid mistake here. But I can't find it now, maybe you could find something... Going home and will look again at it tomorrow.

matrixik:
OK, fixed one simple stupid bug and now it's working.

matrixik:
Hi @timothyb89 keystone-init change ready to review.

keystone-init/README.md
```diff
@@ -174,6 +182,38 @@ containers:
...
+```yaml
+endpoints:
+  example:
+    type:
+    region:
+    url:
+    description:
+    interfaces: []
+```
+In this case service with `example` name will be created first and then
```
kornicameister: So ```example``` is not a name of an endpoint, but rather the name of service. If the service does not exists, it will be created prior to the endpoints creation ? 
Plus looking at other fields like ``type``, `region` it is more like creating a service and assigning the endpoints to it. 
Therefore I don't really like the ```endpoints``` as the way to have it inside the ```preload.yml```. I'd really go with ```services``` and have (optionally) ```interfaces``` changed to ```endpoints```. 

All that is based on a fact that you need a service to assign endpoints to, right ? Therefore the approach you took simse little bit unnatural :skull:

keystone-init/keystone_init.py
```diff
@@ -733,9 +825,35 @@ def load_domains(ks, domains, member_role_name):
 
 
 def load_endpoints(ks, endpoints):
```
kornicameister: Having that said, I'd simply rename this method to load_services.

keystone-init/README.md
kornicameister: @timothyb89 @matrixik told me that my idea is a bit different than the approach you initially had. You think my idea is valid or would you rather stay with the one that @matrixik has proposed ?

timothyb89: I think I wrote the original keystone/preload.yml without a very good understanding of keystone's resource model :smiley: 

It looks like you're right, a [`Service`](https://docs.openstack.org/python-keystoneclient/latest/api/keystoneclient.v3.html#keystoneclient.v3.services.Service) contains some number of [`Endpoints`](https://docs.openstack.org/python-keystoneclient/latest/api/keystoneclient.v3.html#keystoneclient.v3.endpoints.Endpoint), and `interface` is just a property of `Service`. I'd be in favor of changing those names as you suggested.

matrixik: Done

keystone-init/keystone_init.py
matrixik: Done

keystone-init/README.md
```diff
@@ -174,6 +182,38 @@ containers:
...
+`description` is optional.
+
+The `endpoints` block accepts a list of objects. As an example:
+```yaml
+services:
+  example:
+    # ...
+    url: http://top-url:8070/v2.0
+    endpoints:
+      - name: internal
```
kornicameister: In overall, there are 3 possible endpoints **admin**, **internal**, **public**. So having a ```name``` as a word to depict what kind of endpoint we want to create is a bit misleading. I'd roll with ```type``` and maybe add some validation inside the python code to disallow other strings.

matrixik: Now code follow Keystone logic and naming convention.

kornicameister: Peachy ;-)

kornicameister:
I think I prefer being explicit when it comes to defining each endpoint properties:

* url
* interface

etc

defined with each endpoint. There will be a little bit more writing but at least this will prevent unintentional bugs from happening.  

Otherwise the change looks nice.

kornicameister:
And it works :D, just a little nit-pick and we're ready to go.

PR #197, opened 2017-09-08
Rebuild images

knakazono:
!push kibana
!push monasca-notification


knakazono:
@matrixik @timothyb89 @witekest @haruki-yamanashi

PR #191, opened 2017-09-01
Add linters to Travis

matrixik:
How to add Shellcheck: https://github.com/koalaman/shellcheck/wiki/TravisCI

PR #199, opened 2017-09-08
Fix some linter warnings

matrixik:
Some fixes for #191

### DIFF
--- a/elasticsearch-init/upload.sh
+++ b/elasticsearch-init/upload.sh
@@ -4,7 +4,8 @@ TPL_DIR=/templates
 
 _get_tpl_name_from_file() {
     local tpl=$1
-    local tpl_name=$(basename $tpl)
+    local tpl_name
+    tpl_name=$(basename $tpl)
     echo $tpl_name
 }
 
@@ -23,5 +24,3 @@ for template in $TPLS; do
     curl -XPUT --retry 2 --retry-delay 2 $ELASTICSEARCH_URI/_template/${tpl_name} -d @$TPL_DIR/$template
 
 done
-
-
--- a/elasticsearch-init/upload.sh
+++ b/elasticsearch-init/upload.sh
@@ -5,8 +5,8 @@ TPL_DIR=/templates
 _get_tpl_name_from_file() {
     local tpl=$1
     local tpl_name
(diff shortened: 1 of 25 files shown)


matrixik:
Shellcheck is a little less noisy now.
And show some sh/bash specific problems like (we use `#!/bin/sh` everywhere):
```
In monasca-python/build.sh line 48:
    source /clone.sh
    ^-- SC2039: In POSIX sh, 'source' in place of '.' is undefined.
In monasca-python/build.sh line 52:
        local repo=$1
        ^-- SC2039: In POSIX sh, 'local' is undefined.
In mysql-init/init.sh line 139:
    if [ "$d_major" -le "$c_major" -a "$d_minor" -le "$c_minor" -a "$d_patch" -le "$c_patch" ]; then
                                   ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
                                                                ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
```

timothyb89:
Hmm. I think we're going to need to reevaluate some CI decisions. This last build took 7 hours (!) https://travis-ci.org/monasca/monasca-docker/builds/273282923?utm_source=github_status&utm_medium=notification

I'm guessing some networking issue was involved, but it really isn't properly debuggable given how noisy the logs are.

Short term this patch triggered too many container rebuilds, which takes more time and amplifies log output (timothyb89/dbuild#3). If we split this patch up to fix only a couple of modules at a time it should be able to pass CI.

Long term we need to fix dbuild's logging issue and figure out why these builds are taking so long and possibly also put a hard limit on the number of modules we'll build in a single patchset. I've filed #200 to track this.

FWIW this patch really isn't the cause of the problem but unfortunately I think we'll need to work around it while other things are cleaned up.

matrixik:
Ups, sorry for this. I'll split it to smaller PRs.

Hmmm, shouldn't person that open PR get mail from Travis that build failed?

timothyb89:
Yeah, you should have gotten a message... I guess we broke Travis in more ways than one :smiley:

matrixik:
Hmm, no mails with CI errors on my end from any PR...

PR #204, opened 2017-09-11
Fix some linter warnings, part 4.1

matrixik:
Related #191


storm/clean_externals.py
```diff
@@ -34,17 +35,18 @@ def main():
 
     invalid = keep.difference(externals)
     if not keep.issubset(externals):
-        print('Invalid values for KEEP_EXTERNALS: %r' % invalid,
-              file=sys.stderr)
+        print('Invalid values for KEEP_EXTERNALS: {!r}'
```
timothyb89: this seems to be duplicated in #216?

matrixik:
After merging #232 all changes here are duplicated in #216 , my bad.

PR #205, opened 2017-09-11
rebuild monasca-agent-forwarder and monasca-agent-collector

timothyb89:


timothyb89:
must be merged with:

```
!push monasca-agent-forwarder
!push monasca-agent-collector
```

PR #211, opened 2017-09-12
Travis ci gcp buckets

aadams19:
Takes advantage of the --build-log-dir flag in dbuild and saves all build logs to files. Then uploads those files to a GCP bucket.


kornicameister:
I wonder if it wouldn't make more sense to zip all the logs and upload single tarball containing all the logs necessary.

kornicameister:
Also, does it make any sense to upload anything if the build & tests were fine ?

aadams19:
Out plan is to eventually make all of the files browsable, so one big zip wouldn't work for that, but we do want to zip larger individual files and change the type so that the browser does all decoding automatically. We also do want to upload files on success because we are currently seeing instances where all tests pass, but there is a problem. If we didn't upload files, we might lose valuable log data for problems that automatic testing couldn't catch.

ci.py
```diff
@@ -190,9 +190,9 @@ def upload_files(log_dir, bucket):
         file_path = log_dir + f
         if os.path.isfile(file_path):
             if os.stat(file_path).st_size > MAX_RAW_LOG_SIZE:
-                with gzip.open(log_path + '.gz', 'wb')as f_out, open(file_path, 'rb') as f_in:
+                with gzip.open(file_path + '.gz', 'wb')as f_out, open(file_path, 'rb') as f_in:
```
timothyb89: should there be a space between `)` and `as`?

ci.py
```diff
@@ -405,7 +414,7 @@ def update_docker_compose(modules, pipeline):
     compose_dict['version'] = '2'
 
     try:
-        with open(CI_COMPOSE_FILE, 'w') as docker_compose:
+        with open(CI_COMPOSE_FILE, 'wb') as docker_compose:
```
timothyb89: why do these need the `b` flag?

aadams19: the b flag makes it Windows compatible, and having it or not having it has no influence on Linux versions, so it isn't needed at all.

PR #213, opened 2017-09-14
Make Grafana admin user configurable

haruki-yamanashi:
The initialization file for grafana was not configurable by using
enrironmental variables during the startup of the container. This commit
makes it possible to configure Grafana admin user name and password of
the file.

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -11,13 +11,16 @@ arg MONASCA_APP_BRANCH=master
 # `docker build`
 arg REBUILD=1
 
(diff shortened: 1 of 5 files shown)


timothyb89:
@matrixik I agree, we should keep them separate from Grafana's [built-in variables](http://docs.grafana.org/installation/configuration/)

timothyb89:
I think this looks good, but if we want to publish a new image `build.yml` needs to be updated with an incremented version (`-1.1.1` to `-1.1.2` or so). Otherwise, feel free to merge.

matrixik:
Version bumped in #215

PR #216, opened 2017-09-18
Fix some linter warnings, part 5

matrixik:
Separate fixes for mysql-init and storm containers.


storm/clean_externals.py
```diff
@@ -34,17 +35,18 @@ def main():
 
     invalid = keep.difference(externals)
     if not keep.issubset(externals):
-        print('Invalid values for KEEP_EXTERNALS: %r' % invalid,
-              file=sys.stderr)
+        print('Invalid values for KEEP_EXTERNALS: {!r}'
+              .format(invalid, file=sys.stderr))
```
timothyb89: `file=` should be outside the inner parens ... also not totally sure why '%' is preferred?

digitalfishpond:
Some fixes for #191

PR #218, opened 2017-09-18
Fix some linter warnings, part 6

matrixik:


kafka-init/create_topics.py
```diff
@@ -96,7 +96,7 @@ def list_topics():
 
 def create_topic(name, partitions, replicas, configs=None):
     if configs:
-        arg_pairs = map(lambda item: ['--config', '%s=%s' % item],
+        arg_pairs = map(lambda item: ['--config', '{0}={0}'.format(item)],
```
timothyb89: `item` is a tuple with 2 values, '%' expands this but `.format()` will not... we don't want `'{0}={0}'.format(item)` but rather `'{0}={1}'.format(*item)`

(though to be honest I'm not sure of the utility of converting all `%` formatting to use `.format()` anyway...)

matrixik:
Thx @timothyb89 for you comments.
@digitalfishpond will be looking into this lint problems as an starting point to learn more about this project.

digitalfishpond:
some fixes for #191

PR #220, opened 2017-09-18
Use monasca/python for the monasca-alarm image

kornicameister:
monasca/python provides a foundation for the Python
monasca components and several features & fixes that
concern and affects all Python applications.

!push monasca-alarms


kornicameister:
Extracted from #143

kornicameister:
I believe it should feature the **monasca/client** version in _build.yml_ file instead of directly in Dockerfile.

kornicameister:
Will update the PR tommorow

monasca-alarms/build.yml
```diff
@@ -3,3 +3,4 @@ variants:
   - tag: 1.1.2
     aliases:
       - :latest
+  - tag: 1.1.1
```
matrixik: Why this tag is added?

monasca-alarms/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM alpine:3.5
+ARG MON_CLIENT_VERSION="1.6.0"
```
matrixik: There is newer version on master.

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -582,7 +582,7 @@ def main(args=None):
     }
 
     if not monascaclient_found:
-        print("python-monascaclient >= 1.0.9 is required", file=sys.stderr)
+        print("python-monascaclient>=1.6.0<1.7.0 is required", file=sys.stderr)
```
matrixik: Why `<1.7.0` part? And what if newer version is installed?

monasca-alarms/build.yml
kornicameister: Guess rebase gone wrong, I'll bump to ```1.1.3```

monasca-alarms/Dockerfile
kornicameister: Done

monasca-alarms/monasca_alarm_definition.py
kornicameister: Well, the idea is not use untested monascaclient, only the one that is proven to work.

kornicameister: Although, I think I will remove that bit because we have this image on monasca/client which satisfies that condition.

monasca-alarms/Dockerfile
kornicameister: Sorry, but for now we have to stick with `1.6.0`. `1.7.0` comes with breaking changes in the API hence this PR would require too much work to complete. `1.7.0` should be considered as future enhancement.

docker-compose.yml
```diff
@@ -40,6 +40,16 @@ services:
 
   alarms:
     image: monasca/alarms:${MON_ALARMS_VERSION}
+    environment:
```
matrixik: This variables should not be here.
When you set `KEYSTONE_DEFAULTS_ENABLED` in Dockerfile to true they are not needed.

monasca-alarms/Dockerfile
```diff
@@ -1,20 +1,16 @@
...
+ARG MON_CLIENT_VERSION
+FROM monasca/client:${MON_CLIENT_VERSION}
 
 # To force a rebuild, pass --build-arg REBUILD="$(DATE)" when running
 # `docker build`
 ARG REBUILD=1
 
 ENV MONASCA_WAIT_FOR_API=true \
-  KEYSTONE_DEFAULTS_ENABLED=false
+    KEYSTONE_DEFAULTS_ENABLED=false
```
matrixik: I know you did not change it but this variable should be set to `true` like in all other containers and like in monasca-alarms readme...

monasca-alarms/monasca_alarm_definition.py
matrixik: Can monascaclient return it's version so it could be tested?

kornicameister:
@matrixik nope...the client as well as any other component built with ```monasca/python``` is cloned from git with ```---depth 1``` which is messing up the versioning. It always returns ```0.0.1```.

kornicameister:
Thx @aadams19

monasca-alarms/Dockerfile
kornicameister: Ok, will look into it. Thx @matrixik

kornicameister: @matrixik done

kornicameister:
@matrixik everything should be ok now, I applied your suggestion with ```KEYSTONE_DEFAULTS_ENABLED=true``` in last two commits.

PR #222, opened 2017-09-18
Use monasca/python for monasca/client

kornicameister:
monasca/python provides a foundation for the Python
monasca components and several features & fixes that
concern and affects all Python applications.

!push monasca-client


kornicameister:
Extracted from #143

PR #231, opened 2017-09-21
Print logs on failure

aadams19:


ci.py
```diff
@@ -221,6 +231,12 @@ def upload_file(bucket, file_path, file_str=None, content_type='text/plain',
     except Exception as e:
         print ('Unexpected error uploading log files to {}'
                'Skipping upload. Got: {}'.format(file_path, e))
+        if content_encoding == 'gzip':
```
kornicameister: There is no ```with``` here, so I guess the file will stay open, does not really sound ok to me.

aadams19: Good catch. Fixed

PR #219, opened 2017-09-18
Updated monasca-agent images to use monasca/python based monasca/agent-base

kornicameister:
Update images for:

* agent-collector
* agent-forwared

due to changes done for monasca/agent-base.

!push monasca-agent-forwarder
!push monasca-agent-collector

Requires #217 to be merged first.

PR #234, opened 2017-09-28
Configure keystone && fix alarms to work with new python-monascaclient

aadams19:
A recent change to python-monascaclient removed ksclient, so using python-keystoneclient is now necessary to get a token to authenticate with the monasca client.

The keystone variables should only be set if they are not provided so that there is backwards compatibility with keystone v2

### DIFF
--- a/grafana/start.sh
+++ b/grafana/start.sh
@@ -1,4 +1,8 @@
 #!/bin/sh
 
+if [ -n $GF_DATABASE_PORT ]; then
+  export GF_DATABASE_HOST=$GF_DATABASE_HOST":"$GF_DATABASE_PORT
+fi
+
 python /template.py /etc/grafana/grafana.ini.j2 /etc/grafana/grafana.ini
 /go/bin/grafana-server -config /etc/grafana/grafana.ini -homepath /var/lib/grafana
--- a/grafana/start.sh
+++ b/grafana/start.sh
@@ -1,8 +1,4 @@
 #!/bin/sh
 
-if [ -n $GF_DATABASE_PORT ]; then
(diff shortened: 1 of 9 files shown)


monasca-agent-collector/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM monasca/agent-base:master-20170901-210328
+ARG MON_AGENT_BASE_VERSION
```
timothyb89: This won't have any effect, `ARG` needs to be before `FROM` and then needs to actually be used.

Alternatively, we can just update the `FROM` and not have an `ARG` at all, which IMO is a bit simpler and still works fine since we don't have a python3 agent to worry about.

monasca-agent-collector/build.yml
```diff
@@ -3,3 +3,5 @@ variants:
   - tag: master
     aliases:
       - :master-{date}-{time}
+    args:
```
timothyb89: This can be removed if we don't use the `ARG` per the above comment.

monasca-agent-forwarder/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM monasca/agent-base:master-20170901-210328
+ARG MON_AGENT_BASE_VERSION
```
timothyb89: same issue as the collector

monasca-agent-collector/Dockerfile
timothyb89: actually I just merged https://github.com/monasca/monasca-docker/pull/219 which will cover this as well

timothyb89:
The Grafana change was moved to #234

PR #235, opened 2017-09-28
Build grafana url in the env var

aadams19:


timothyb89:
I think the build issue might be fixable by using `yarn --no-emoji` in the Dockerfile: https://github.com/monasca/monasca-docker/blob/master/grafana/Dockerfile#L29

The real solution would be to support utf8 in dbuild, but that is probably not a quick fix.

timothyb89:
Turns out utf-8 support in dbuild was not too difficult (https://github.com/timothyb89/dbuild/commit/a206ffb524e5c70431e2ca328ec1cce908ccd3d8) :smile: 

Should be alright now.

PR #240, opened 2017-10-03
Add retention to database

rbrndt:
This only adds the retention at creation. If the database already exists, the retention settings will not be applied.


kornicameister:
I think we should push the new image version, modify the **build.yml** to bump it, and also annotate the commit msg with ```!push influxdb-init```

timothyb89:
slight addendum: we push automatically when build.yml is updated, so `!push` would not be needed in this case :smiley:

timothyb89:
Also, is there some value for `INFLUXDB_RETENTION` that keeps current behavior? i.e. keep forever? Or would we just use a really large value in that case?

rbrndt:
Hmm, that's a good point. I was hoping to add as little bash as possible, but I should probably allow not specifying.

kornicameister:
@rbrndt could you add just a small note to README.md as to to the purpose of variables included in setting the retention ?

matrixik:
Hm, `influxdb-init` don't have README so you need to create it, for now just add info about this new variables to it. No long description needed.

influxdb-init/README.md
```diff
@@ -0,0 +1,8 @@
+# Influx Init
+
+# Variables
```
matrixik: Should be second level so `##`

influxdb-init/README.md
```diff
@@ -0,0 +1,8 @@
...
+INFLUXDB_DEFAULT_RETENTION - determines how long the data should be stored before it is dropped (default for all series created on the database) https://docs.influxdata.com/influxdb/v1.1/query_language/database_management/#retention-policy-management
```
matrixik: Put variables like INFLUXDB_DEFAULT_RETENTION in back ticks `INFLUXDB_DEFAULT_RETENTION`.
Too long lines, split them around 80-100 column, please.
Put long links at the end of file and link to them from here (you could check how to do it in [Kafka-init README](https://raw.githubusercontent.com/monasca/monasca-docker/master/kafka-init/README.md) but you could use aliases instead of numbers).
We are using [InfluxDB 1.3](https://github.com/monasca/monasca-docker/blob/master/.env#L1) so link to the same version, please.

influxdb-init/README.md
```diff
@@ -0,0 +1,8 @@
...
+Acceptable values are https://docs.influxdata.com/influxdb/v1.1/query_language/spec/#durations and 'INF' for infinite (Default: INF)
+
+INFLUXDB_SHARD_DURATION - determines the size of data shards. An entire shard is dropped once all of the data it contains is out of the retention period.
```
matrixik: Same comments.

rbrndt:
I see influx_init_version set to latest, do we still need it to be the new version?

kornicameister:
Nope, that's fine as it is I believe.

PR #242, opened 2017-10-04
Make SSL configurable in Storm and Thresh

aadams19:


storm/templates/thresh-config.yml.j2
```diff
@@ -121,11 +121,11 @@ sporadicMetricNamespaces:
 
 database:
   driverClass: com.mysql.jdbc.Driver
-  url: "jdbc:mysql://{{ MYSQL_DB_HOST | default('mysql') }}:{{ MYSQL_DB_PORT | default('3306') }}/{{ MYSQL_DB_DATABASE | default('mon') }}?useSSL=true"
+  url: "jdbc:mysql://{{ MYSQL_DB_HOST | default('mysql') }}:{{ MYSQL_DB_PORT | default('3306') }}/{{ MYSQL_DB_DATABASE | default('mon') }}?useSSL={{ USE_SSL_ENABLED }}"
   user: "{{ MYSQL_DB_USERNAME | default('thresh') }}"
   password: "{{ MYSQL_DB_PASSWORD | default('password') }}"
   properties:
-      ssl: false
```
craigbr: Makes you wonder which way it used to be configured for ssl? false from here or true from line 124?

PR #245, opened 2017-10-06
Keystone client bug

aadams19:
This container should be compatible with python-monascaclient v1.7.0


kornicameister:
To be fair, the `1.7.0` should on its own take care of recognizing all ```OS_``` variables hence the parsing done in the script prior to execution shouldn't be necessary - but I guess that only happens in `shell` mode.

Nevertheless, I don't really feel like retrieving the token is a must-have as we can rely on [this](https://github.com/openstack/python-monascaclient/blob/master/monascaclient/client.py#L149) and basically allow client to do its job. I would skip retrieving the token and simply pass the `username` and `password`

aadams19:
Good catch. I took out all keystoneclient logic that duplicated code getting a token.

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -207,9 +184,15 @@ def run(self, data_file):
 
         yaml_data = yaml.safe_load(yaml_text)
 
-        self._keystone_auth()
-        self._monasca = client.Client(self._args['api_version'], self._api_url, token=self._token)
-        self._print_message('Using Monasca at {}'.format(self._api_url))
+        if self._args['monasca_api_url'] is None:
```
kornicameister: That's a bit incorrect, but not here but in client - wonder if shouldn't support endpoint retrieving in client. But that's a note that does not change anything in here.

kornicameister:
You saw the error [here](https://storage.googleapis.com/monasca-ci-logs/monasca-docker%2F2017-10-10_23-19-46_monasca_logs%2Frun%2Fdocker_log_monascadocker_alarms_1.log.gz). Guess something is bit wrong with the alarms.

matrixik:
If some element fail it should make travis run to fail but all of them still are green. Maybe saving logs to Google Cloud Storage somehow suppress errors.

aadams19:
Good catch on the failure. There is a fix for this ci issure at https://github.com/monasca/monasca-docker/pull/251

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -441,6 +423,10 @@ def _get_parser():
                         default=_env('OS_PROJECT_ID'),
                         help='Defaults to env[OS_PROJECT_ID].')
 
+    parser.add_argument('--os-tenant-name',
```
kornicameister: I am not so sure about that, `tenant-name` is basically deprecated in favor of `project-name`. That said, I think we shouldn't introduce that but remove it.

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -564,26 +555,34 @@ def main(args=None):
...
+        'service_type': args.os_service_type,
         'insecure': args.insecure,
         'monasca_api_url': args.monasca_api_url,
         'api_version': args.monasca_api_version,
         'verbose': args.verbose
     }
 
     if not monascaclient_found:
-        print("python-monascaclient>=1.6.0<1.7.0 is required", file=sys.stderr)
+        print("python-monascaclient>=1.6.0 is required", file=sys.stderr)
```
kornicameister: I think we should change that now into `>=1.7.0`. Or even actually remove that to be honest. `monasca-alarm` image is based on top of `monasca-client` image meaning that client will be always available.

kornicameister:
One bit is to increase the version of the client inside the `build.yml`.

monasca-alarms/monasca_alarm_definition.py
aadams19: done

monasca-alarms/monasca_alarm_definition.py
aadams19: done

aadams19:
The cli is forced to use keystone v2 because of a default value for OS_IDENTITY_API_VERSION, and the python client is hardcoded to use v3. Since this container uses both, this must be remedied upstream.

aadams19:
It looks like an upstream fix to some of our problems is @ https://review.openstack.org/#/c/512648

I filed a bug report for the other major problem that this container is having @ https://storyboard.openstack.org/#!/story/2001246

witekest:
@aadams19 python-monascaclient 1.8.0 has been released with fix for https://storyboard.openstack.org/#!/story/2001158 on master.

monasca-alarms/build.yml
```diff
@@ -1,7 +1,7 @@
 repository: monasca/alarms
 variants:
-  - tag: 1.1.3
+  - tag: 1.1.4
     aliases:
       - :latest
     args:
-      MON_CLIENT_VERSION: "1.6.0"
+      MON_CLIENT_VERSION: "1.7.0"
```
kornicameister: Let's change the client version to **1.8.0** here :)

aadams19: It seems this isn't going to work quite yet: https://hub.docker.com/r/monasca/client/tags/

manifest for monasca/client:1.8.0 not found

matrixik: Looks like tagging 1.8 failed:
https://travis-ci.org/monasca/monasca-docker/jobs/295171347

matrixik: And after merging #313 https://travis-ci.org/monasca/monasca-docker/jobs/295486366
https://storage.googleapis.com/monasca-ci-logs/monasca-docker%2F2017-11-01_15-26-15_monasca_logs%2Fbuild%2F2017-11-01-15-26-16-monasca-client-1.8.0.log.gz
```
+ getopt -n parse-options -o r:b:e:d:c:u:q: -- -r https://git.openstack.org/openstack/python-monascaclient -b 1.8.0 -q master
+ OPTS= -r 'https://git.openstack.org/openstack/python-monascaclient' -b '1.8.0' -q 'master' --
+ echo Failed parsing options.
Failed parsing options.
+ exit 1
```

aadams19: It is probably because we need https://travis-ci.org/monasca/monasca-docker/builds/295486362 to pass, should go through soon though

aadams19:
Progress!
But this review is still blocked by https://storyboard.openstack.org/#!/story/2001246

witekest:
@aadams19 I guess you need something like

```
OS_IDENTITY_API_VERSION=${OS_IDENTITY_API_VERSION:-"3"}
```

around [this line](https://github.com/monasca/monasca-docker/blob/master/monasca-alarms/start.sh#L13).

timothyb89:
I just tested using `OS_IDENTITY_API_VERSION=3` and it worked perfectly! I think we're good to merge this now.

aadams19:
That's exciting! Do we need any changes, or are we good?

timothyb89:
Just needed a version bump (just pushed it) and I think we're good to go.

aadams19:
Thanks @witekest @kornicameister and @timothyb89 !!!

timothyb89:
darn, still hitting https://storyboard.openstack.org/#!/story/2001246

I think a better solution will be to just add the wait loop to `monasca_alarm_definition.py` rather than calling out to the monasca-client in bash

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -544,25 +616,28 @@ def main(args=None):
         'domain_id': args.os_domain_id,
         'domain_name': args.os_domain_name
     }
-    ks_kwargs = dict((k, v) for k, v in all_keystone_kwargs.iteritems() if v)
+    ks_kwargs = {k: v for k, v in all_keystone_kwargs.iteritems() if v}
+    session = get_keystone_client(ks_kwargs)
 
     kwargs = {
-        'keystone_kwargs': ks_kwargs,
+        'keystone_'
```
aadams19: Should this have a value?

timothyb89: oops, yup

monasca-alarms/monasca_alarm_definition.py
```diff
@@ -541,6 +518,71 @@ def _env(*vars, **kwargs):
...
+    if not api_url:
+        raise Exception('Error: monasca_api_url is required')
+
+    monasca = client.Client(args['api_version'], api_url, **keystone_args)
+    print('Using Monasca at {}'.format(api_url))
+
+    for i in range(MONASCA_RETRIES):
+        try:
+            # just make an api call that should always succeed
+            monasca.notifications.list(limit=1)
```
aadams19: I'm still confused about why we can't just let the monascaclient authenticate to keystone. It seems like get_keystone_client does a whole bunch of work that is just repeated in the monascaclient. Why do we need to authenticate to keystone explicitly?

timothyb89: Yeah, originally I was hoping not to touch the `MonascaLoadDefinition` class at all, so it made more sense to have that in a separate step. Since I had to do that anyway, they probably could be combined now.

PR #253, opened 2017-10-13
Add alarm definition controller dockerfile

timothyb89:


alarm-definition-controller/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM alpine:3.6
+
+RUN apk add --no-cache ca-certificates tini
+
+COPY --from=go-builder \
+    /go/src/github.com/monasca/alarm-definition-controller/adc \
+    /adc
+
+ENTRYPOINT ["/sbin/tini", "--"]
+CMD ["/adc"]
```
mhoppal: is this really the binary name?

timothyb89: it is since I named it that :smiley: 

the alternative was `alarm-definition-controller`, so...

PR #254, opened 2017-10-17
Update version of monasca/storm for monasca/thresh

kornicameister:
Make monasca/thresh inherit from monasca/storm:1.1.1-1.0.0.


timothyb89:
I feel like `test_create_metric_with_multibyte_character` is really unreliable for some reason. It seems to  be the only test to fail most of the time...

kornicameister:
Yeah, wonder how we should proceed with this now ;/. It's a bit funny because CI in review.openstack.org passes all the time with Storm 1.1.1

matrixik:
Will be fixed in https://github.com/monasca/monasca-docker/pull/345

Issue #225, opened 2017-09-18
Additional log upload improvements


timothyb89:
#211 adds log uploads to a GCS bucket. We can make a few more small improvements that would improve quality of life a lot:

- [ ] Compress (gzip) logs over a certain size, say 10 KiB. Files should be named `.log.gz` versus `.log` and have their Content-Encoding set to `gzip`.
- [ ] Add a small dashboard for each run using the manifest file. We could provide some basic info about the build and link directly to each container log.
- [ ] Look at including docker's json logs as well. They include timestamps and the stream name for each message so we could have a nice preview in the browser.

timothyb89:
An easier way to get timestamps in logs: add `-t` to the `docker logs` call

aadams19:
See https://github.com/monasca/monasca-docker/pull/211/commits/72e5ee14cdc7021fe41b57e9c610bf10bb6da832 for compress and timestamp fixes

kornicameister:
@aadams19 do you want to implement all those points in #211 ?

aadams19:
No, our plan is just do those two changes in https://github.com/monasca/monasca-docker/commit/72e5ee14cdc7021fe41b57e9c610bf10bb6da832 for now since they were fairly simple changes with high rewards. The rest we will have to revisit sometime in the future

timothyb89:
So it looks like we failed to account for situations where secure environment variables aren't available, that is, any PR from a branch that isn't this repo (i.e. a fork, where most PRs should come from). For example: https://travis-ci.org/monasca/monasca-docker/jobs/277478462#L649

Turns out this is a fairly significant issue (really always the case with 3rd party CI). I think the quick solution is:
 * Use travis's JWT addon: https://docs.travis-ci.com/user/jwt
 * Make a tiny log ingester on Cloud Functions to take 1+ logs and a JWT. Post logs to that endpoint instead of uploading to GCS directly.

PR #256, opened 2017-10-17
Fix some linter warnings, part 7

digitalfishpond:
fixes some lint errors for #191


timothyb89:
Hmm, we might need to split this patch up a bit more, it's generating too much log output and getting killed by travis. Unfortunately we'll need something like https://github.com/monasca/monasca-docker/issues/225#issuecomment-330675960 to avoid this in the future.

digitalfishpond:
Pull request has been split into 3 smaller PRs.

PR #258, opened 2017-10-18
Fix some linter warnings, part 7b

digitalfishpond:
splitting #256 to avoid log overload


digitalfishpond:
Agreed with @kornicameister that after all and henceforth, would be best to make one big pull request at the end of all linter resolution PRs updating .env and build.yaml files to avoid systematic conflicts on the .env file.

digitalfishpond:
redundant PR: fixes split between #277 #276 #278

PR #259, opened 2017-10-18
Fix some linter warnings, part 7c

digitalfishpond:
splitting #256 to avoid log overload


digitalfishpond:
@kornicameister adding .env to every PR generates too many logs again, causing Travis to fail (the reason  for making many small PRs), so I will make additional PRs at the very end updating env and build.yml files

digitalfishpond:
will bump kafka-init version in separate pr to avoid errors

PR #261, opened 2017-10-18
Use a temporary branch for the persister to fix config error

timothyb89:
monasca-persister started treating the `ip_address` config option as
a strict IP address and failed validation when we passed it a
hostname.

For now we'll switch the persister to use a temporary branch while we
wait for the fix to merge.


timothyb89:
side note: we need to make sure to update `.env` when this merges, we didn't catch this because we're still testing against a build from June 😃

kornicameister:
@timothyb89 @mhoppal Should help you out ;-) => https://review.openstack.org/513259

kornicameister:
Already merged => https://github.com/openstack/monasca-persister/commit/66fe6f98c079490d2d5de4c161da1d8b3801cda4

timothyb89:
Thanks, will just update the persister now :smiley:

PR #262, opened 2017-10-18
Only create mysql stuff if it doesn't already exist

aadams19:
If the table or database already exists, that shouldn't throw an error and the pod should be allowed to exit gracefully.


timothyb89:
`MYSQL_INIT_VERSION` in `/.env` should be bumped too.

timothyb89:
Actually, scratch that. I have a bit patch of updates to that file, I'll include this patch.

mysql-init/init.sh
```diff
@@ -103,7 +103,7 @@ schema_upgrade() {
   # ash doesn't support arrays, this seems to be the most concise way to get
   # fields by index
   set "$version"
-  if [ "$#" -ne "3" ]; then
+  if [ "$#" -ne "1" ]; then
```
timothyb89: Fairly sure this should stay at 3, could be wrong though.

timothyb89: Ahhh. Found the issue. There should be no quotes in `set "$version"` (changed in https://github.com/monasca/monasca-docker/commit/d03d2a39398b071859ec5d414a2e3ff17dbb9c11#diff-ecae5aa2959527c258187b0881107733L105)

We will need to add a shellcheck exception like on line 127, though.

PR #264, opened 2017-10-19
Add linters to Travis: Python

digitalfishpond:
Splitting #191 for manageability

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,17 @@ before_install:
   - pip install pip --upgrade
   - pip install git+https://github.com/timothyb89/dbuild.git
   - pip install google-cloud-storage
+  - pip install flake8 flake8-blind-except flake8-builtins flake8-pep3101 flake8-coding
+  - pip install pylint
   - docker --version
   - docker-compose --version
 
 jobs:
   include:
+    - stage: linting-flake8
+      script: flake8
+    - stage: linting-pylint
+      script: pylint
     - stage: metrics-pipeline
       script: python ci.py metrics
     - stage: logs-pipeline


matrixik:
I think we should configure it with same/similar configuration like in openstack 
E.g. max line length is 120 in monasca-api: https://github.com/openstack/monasca-api/blob/master/tox.ini#L147

matrixik:
I think that most standard/popular way of having configuration for Python would be `setup.cfg` file
http://renesd.blogspot.com/2017/02/setupcfg-solution-to-python-config-file.html

Some more info for consideration with pylint:
https://codewithoutrules.com/2016/10/19/pylint/

.travis.yml
```diff
@@ -21,13 +21,21 @@ before_install:
...
 
 jobs:
   include:
     - stage: lint-shellcheck
       script: bash -c 'shopt -s globstar; shellcheck **/*.sh'
+    - stage: linting-flake8
+      script: flake8
+    - stage: linting-pylint
+      script: pylint --rcfile=setup.cfg *
+    - stage: lint-shellcheck
```
matrixik: Remove duplicated shellcheck stage

setup.cfg
```diff
@@ -0,0 +1,21 @@
+[flake8]
+# D203  1 blank line required before class docstring
+# H201  no 'except:' at least use 'except Exception:'
+# H302  import only modules
+# H405  multi line docstring summary not separated with an empty line
+ignore = D203,H201,H302,H405
+# H106: Don’t put vim configuration in source files
+# H203: Use assertIs(Not)None to check for None
+enable-extensions = H106,H203
```
matrixik: All `Hxxx` are comming from OpenStack `hacking` https://docs.openstack.org/hacking/latest/ so it also need to be installed

setup.cfg
```diff
@@ -0,0 +1,21 @@
...
+max-complexity = 50
+max-line-length = 120
+exclude =.env,.git,doc,*lib/python*,build,dist,*.egg
+show-source = True
+builtins = _
+
+[pylint]
+disable = C0103,C0111,F0010
```
matrixik: Put comments with description what are you disabling (like in 2-5 lines)

.travis.yml
digitalfishpond: done

setup.cfg
digitalfishpond: done

setup.cfg
digitalfishpond: done

PR #263, opened 2017-10-19
Add linters to Travis: Scripts

digitalfishpond:
splitting #191 for manageability

PR #265, opened 2017-10-19
Fix some linter warnings, part 9

digitalfishpond:
fixes some errors for #263 (formerly #191)
!push kibana


kibana/build.yml
```diff
@@ -2,7 +2,7 @@ repository: monasca/kibana
 variants:
   - tag: latest
     aliases:
-      - :4.6.3-master
+      - :4.6.4-master
```
kornicameister: That is actually not true, this particular case is a Kibana version itself set to be 4.6.3. That said, I guess you need to have
```!push kibana``` to update master image directly.

@timothyb89 is that true ?

PR #269, opened 2017-10-19
Rebuild the persister

timothyb89:
We need to pick up https://review.openstack.org/#/c/513259


timothyb89:
Need `!push monasca-persister-python` on the merge commit

PR #271, opened 2017-10-19
Update agent-collector and agent-forwarder to use new base image

timothyb89:


timothyb89:
side note: no `!push` needed here

PR #274, opened 2017-10-20
Change log-agent conf ext from p2 to j2

matrixik:


matrixik:
Fixed PR title to `Change log-agent conf ext from p2 to j2`

PR #275, opened 2017-10-20
Fix some linter warnings, part 7b1

digitalfishpond:
further splitting #258 and bumping kafka-init version for #259


digitalfishpond:
splitting this further

PR #290, opened 2017-10-23
Update thresh's base storm image

timothyb89:


timothyb89:
Still on storm `1.0.3` for now, though I suppose we can cut over to `1.1.0` if/when that's been validated.

PR #297, opened 2017-10-24
Fix some linter warnings, part 21

digitalfishpond:
fixes some errors for #263 (formerly #191)
!push monasca-thresh

PR #298, opened 2017-10-25
Check if #297 fails due to code changes

digitalfishpond:
DO NOT MERGE!
Just checking to see if Travis still fails with nothing but an extra line in the monasca-thresh module


digitalfishpond:
Proof only. Not intended to be merged

PR #310, opened 2017-10-30
Fix Python linter errors - Part 1

digitalfishpond:
Fixes Python lint errors for #264


digitalfishpond:
@kornicameister  @timothyb89 Travis is failing to start. This was passing all checks earlier this morning, I re-started Travis with no changes and it now fails.

ci.py
```diff
@@ -438,9 +438,9 @@ def update_docker_compose(modules, pipeline):
     try:
         with open(CI_COMPOSE_FILE, 'w') as docker_compose:
             yaml.dump(compose_dict, docker_compose, default_flow_style=False)
-    except:
+    except (RuntimeError, IOError):
```
matrixik: https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

ci.py
```diff
@@ -449,7 +449,7 @@ def load_yml(yml_path):
         with open(yml_path) as compose_file:
             compose_dict = yaml.safe_load(compose_file)
             return compose_dict
-    except:
+    except(RuntimeError, IOError):
```
matrixik: Use `EnvironmentError` instead of `IOError`
https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError
and catch also error from yaml

PR #311, opened 2017-10-30
Add tag for python-monascaclient 1.8

aadams19:


kornicameister:
Ok, let's update the alarms in seperate PR to use new client

monasca-client/build.yml
```diff
@@ -1,14 +1,24 @@
 repository: monasca/client
 variants:
-  - tag: 1.7.0
+  - tag: 1.8.0
     aliases:
       - :latest
+      - :1.8
+      - :1
+    args:
+      CLIENT_BRANCH: 1.7.0
```
kornicameister: Whoa...shouldn't this be 1.8.0 ?

aadams19: Ooops, good catch

PR #314, opened 2017-11-01
Fix getopt call in monasca-python/build.sh

timothyb89:
The change made the getopt call check for success instead of failure.


timothyb89:
I'm going to go ahead and merge now so we can get a new client container building, feel free to ping me / revert in case of any issues.

digitalfishpond:
Thanks for this @timothyb89

PR #317, opened 2017-11-02
Add git to monasca/client build

timothyb89:
git was missing from the image causing the client build to fail.


timothyb89:
In this week's episode of "How Did This Ever Work"...

timothyb89:
Also, we'll need `!push monasca-client` in the merge commit message.

PR #321, opened 2017-11-06
Allow kafka container to stay alive after kafka exits

craigbr:
This will allow debugging of error conditions and handling problems
like a full disk.


craigbr:
Version bumped

PR #322, opened 2017-11-06
Add the support for upgrading kafka topic configs

mhoppal:


kafka-init/create_topics.py
```diff
@@ -177,10 +205,12 @@ def main():
     existing_topics = list_topics()
     logger.info('Kafka has topics: %r', existing_topics)
 
-    created_topics = create_topics(default_config, existing_topics)
+    created_topics, existing_topics_config = create_topics(default_config, existing_topics)
     logger.info('Topic creation finished successfully. Created: %r',
                 created_topics)
 
+    logger.info(existing_topics_config)
```
rbrndt: This message should probably be more descriptive or removed

timothyb89: Agree, but do we want to address that in a follow-up?

kafka-init/create_topics.py
```diff
@@ -161,13 +156,46 @@ def create_topics(default_config, existing_topics):
...
+        if configs_to_delete:
+            logger.info('Removing topic configuration options %s from topic %s', configs_to_delete, topic)
+            config_input = []
+            for config_opt in configs_to_delete:
+                config_input.append("--delete-config")
+                config_input.append(config_opt)
+            kafka_topics("alter", ["--topic", topic] + config_input)
+
+        if config:
+            logger.info('Adding/Updating topic configuration options %s to topic %s', config.keys(), topic)
```
craigbr: It always calls alter on the topic even if the config options already match. Probably we don't care about that.

kafka-init/create_topics.py
```diff
@@ -161,13 +156,46 @@ def create_topics(default_config, existing_topics):
                 'partitions not set for topic %s, it will not be created!',
                 topic_name)
             continue
-
+        if topic_name in existing_topics:
+            logger.info('Topic %s already exists, ensuring configuration options match up',
```
craigbr: I would move this log statement into update_topic_configs(). That way if something goes wrong, we'll know which topic was being updated

kafka-init/create_topics.py
mhoppal: i can change

kafka-init/create_topics.py
mhoppal: good call :)

kafka-init/create_topics.py
mhoppal: yeah should hvae no harm

PR #323, opened 2017-11-09
Fix Python linter error - Part 2

digitalfishpond:
Fixes Python lint errors for #264


ci.py
```diff
@@ -17,23 +17,21 @@
 #
 
 from __future__ import print_function
-
```
matrixik: Don't remove this empty line

ci.py
```diff
@@ -17,23 +17,21 @@
...
 import datetime
+from google.cloud import storage
```
matrixik: `from ... import ...` should be separated from `import ...`

digitalfishpond:
@matrixik requested changes done

PR #328, opened 2017-11-15
Add Elasticsearch Curator for data retention

matrixik:


matrixik:
Ready for review

elasticsearch-curator/actions.yml
```diff
@@ -0,0 +1,3 @@
+curator_actions: []
```
adriancz: Maybe is good idea to add default curator_action configuration, because when this option is empty list docker image  will return error.

matrixik: Done

PR #329, opened 2017-11-16
Add 'domain_roles' to user object

sjmc7:
Adds 'domain_roles' to the user object which acts similarly to 'roles' but applied
to the domain element of which the user is a child.


sjmc7:
Done @timothyb89

PR #330, opened 2017-11-17
Add prometheus metric-type options to agent-base

rbrndt:


timothyb89:
side note: needs `!push monasca-agent-base` added to the merge commit message

PR #345, opened 2017-11-28
Use new storm container for thresh

craigbr:
Need the UI_PORT to STORM_UI_PORT fix

Need quotes in if statement otherwise we get a syntax error when
trying to run with storm


matrixik:
Merged in  #357

PR #347, opened 2017-11-28
Make collector_restart_interval configurable

craigbr:


monasca-agent-base/agent.yaml.j2
```diff
@@ -35,7 +35,8 @@ Main:
   num_collector_threads: {{ NUM_COLLECTOR_THREADS | default(1) }}
   pool_full_max_retries: {{ POOL_FULL_MAX_TRIES | default(4) }}
   sub_collection_warn: {{ SUB_COLLECTION_WARN | default(6) }}
-  collector_restart_interval: 24
+  autorestart: {{ AUTORESTART | default(True) }}
```
mhoppal: should we default to false? I feel like the defaults should be not to restart

craigbr: That would be a change of behavior. The current default is to restart. But, I can go either way with this.

PR #372, opened 2017-12-08
Refactor job-cleanup

### LINKED ISSUE #45: Cleanup should not delete itself when it encounters problems
The job-cleanup script tries to delete itself when finished. It should only do this when everything is successful, since it will end up covering up failures in init jobs it needs to kill, like thresh-init.

timothyb89:
Several improvements to job-cleanup
 - Clean orphaned job pods (completed job-pods with a missing job)
 - Don't delete cleanup-job pod, only the job (makes debugging
   possible, fixes #45)
 - Reduce default number of retries 24 to 5 (it turns out we usually
   would prefer to just kill things that are still running)
 - Use python 3.5
 - Use tiny-kubernetes instead of embedded k8s client
 - Use python logging


job-cleanup/cleanup.py
```diff
@@ -15,27 +15,31 @@
...
+LOG_LEVEL = logging.getLevelName(os.environ.get('LOG_LEVEL', 'INFO'))
+logging.basicConfig(level=LOG_LEVEL,
+                    format='%(asctime)-15s - %(levelname)-8s - %(message)s')
+logger = logging.getLogger(__name__)
 
 pod_is_self = True
 
 
-def get_current_namespace():
+def get_current_namespace() -> str:
```
aadams19: Why type cast everything now?

timothyb89: They aren't type casts, they're [type hints](https://www.python.org/dev/peps/pep-0484/) for python 3.5. Decent IDEs (pycharm, etc) will warn about improper variable usage with them enabled.

job-cleanup/cleanup.py
```diff
@@ -138,46 +147,78 @@ def label_defunct(client, namespace, job):
...
-    else:
-        selector = 'app={}'.format(app)
+        if pod.status.phase == 'Running':
+            logger.info('Orphaned pod %s is still running and will not be '
+                        'deleted', pod.metadata.name)
+            continue
 
-    jobs = client.get('/apis/batch/v1/namespaces/{}/jobs', namespace,
-                      params={'labelSelector': selector})
+        orphaned.append(pod)
```
aadams19: Why store the whole pod in this list? Couldn't you just have a list of pod names?

timothyb89: That would work fine too, I think both are mostly equivalent here.

PR #379, opened 2017-12-22
Empty commit to create new tempest tests container

craigbr:
New Monasca API requires a new tempest test container as "{}" is
now allowed in dimension names and values

Remember to do a !push tempest-tests in the merge message


craigbr:
tempest tests has an explicit version. Replace with https://github.com/monasca/monasca-docker/pull/380

PR #384, opened 2018-01-04
Hide Monasca password from ps

aadams19:
If you so happened to do a ps -ef while a HEALTHCHECK is running for the API, then you would see a plaintext password. This changes the HEALTHCHECK to use environment variables to avoid that problem.


aadams19:
https://github.com/monasca/monasca-docker/pull/386

PR #386, opened 2018-01-04
Remove API Healthcheck

aadams19:
The healthcheck is causing problems with the passwords being sent over the command line and the keystone token refreshing on every run. The plan is to replace the check in the future with something smarter and more secure


aadams19:
@kornicameister and @witekest Do you guys have any objections to this?

PR #396, opened 2018-01-16
Change default Grafana admin user

matrixik:
If Grafana and Keystone have admin user with same name they will clash
and user in Grafana loss admin status.


matrixik:
Fix for ##188

PR #398, opened 2018-01-17
Use monasca org Grafana app repo in Dockerfile

matrixik:
Proper repo was used only when building in Travis, local builds was
still using old repository.


habeeb-umo:
This change was incorporated into #400

matrixik:
Thank you, we had some problems with communication from Grafana to Monasca, but did not have time to investigate more.

PR #407, opened 2018-02-07
Pull most recent API changes in

aadams19:


aadams19:
Oh no! That's from before https://review.openstack.org/#/c/533007/ . Let me look into why you would be seeing that

PR #409, opened 2018-02-08
Update docker-ce with addons option in Travis

matrixik:


matrixik:
Looks like previous way of updating docker-ce in Travis stopped working after Docker moved from 18.01 to 18.02:
https://travis-ci.org/monasca/monasca-docker/jobs/338911100

Use more standard way of updating it in Travis with `addons` section in config.

PR #419, opened 2018-02-28
Bump InfluxDB-Watcher to include read bug fix

aadams19:
Pulls in https://github.com/monasca/monasca-watchers/pull/12

### DIFF
--- a/influxdb-watcher/build.yml
+++ b/influxdb-watcher/build.yml
@@ -1,5 +1,5 @@
 repository: monasca/influxdb-watcher
 variants:
-  - tag: ':latest'
+  - tag: 'latest'
     aliases:
-      - 0.0.2
+      - :0.0.2
--- a/influxdb-watcher/build.yml
+++ b/influxdb-watcher/build.yml
@@ -1,5 +1,5 @@
 repository: monasca/influxdb-watcher
 variants:
-  - tag: 0.0.1
+  - tag: ':latest'
     aliases:
-      - ':latest'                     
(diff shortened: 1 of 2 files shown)


influxdb-watcher/build.yml
```diff
@@ -1,5 +1,5 @@
 repository: monasca/influxdb-watcher
 variants:
-  - tag: 0.0.1
+  - tag: ':latest'
```
timothyb89: whoops, there shouldn't be a colon here

influxdb-watcher/build.yml
```diff
@@ -1,5 +1,5 @@
...
     aliases:
-      - ':latest'                     
+      - 0.0.2
```
timothyb89: and we do need a colon here

timothyb89:
it's a bit concerning that CI didn't catch that... I would expect an error...

influxdb-watcher/build.yml
aadams19: Done

influxdb-watcher/build.yml
aadams19: Lol, totally mixed them up

timothyb89:
oh!

```
Module influxdb-watcher does not belong to metrics, skipping
```

we need to add influxdb-watcher to the metrics pipeline: https://github.com/monasca/monasca-docker/blob/master/ci.py#L42

that can be a separate commit, as long as we actually publish something here

PR #420, opened 2018-02-28
Add watchers to ci metrics pipeline

aadams19:
The three watchers should be included in the metrics pipeline in ci.


ci.py
```diff
@@ -42,9 +42,12 @@
 METRIC_PIPELINE_MODULE_TO_COMPOSE_SERVICES = {
     'monasca-agent-forwarder': 'agent-forwarder',
     'zookeeper': 'zookeeper',
+    'zookeeper-watcher': 'zookeeper',
```
timothyb89: should this be `'sookeeper-watcher': 'zookeeper-watcher',` ?

aadams19: yes, good catch

aadams19:
It looks like this is failing ci because the watchers aren't all in docker-compose.yaml . I'll get a fix up for that asap

PR #422, opened 2018-04-10
[aggregator] Move config files to subdirectories

timothyb89:
Due to a recent Kubernetes CVE, subpath mounting (which allowed
single-file mounts to `/`) is broken. As the old aggregator image
expected to find its config files in `/`, it won't run in affected
versions.

This adds a workaround to `start.sh` so `config.yaml.j2` is copied
from `/config` and `aggregation-specifications.yaml` is copied from
in `/specs`. This allows ConfigMaps to be mounted to these individual
directories without any trouble.

As this is a breaking change to the aggregator image, this is a new
minor version release.

### DIFF
--- a/monasca-aggregator/Dockerfile
+++ b/monasca-aggregator/Dockerfile
@@ -5,7 +5,7 @@ arg AGGREGATOR_BRANCH=master
 
 env GOPATH=/go
 
-run apk add --no-cache openssl py2-jinja2 && \
+run apk add --no-cache openssl py2-jinja2 librdkafka && \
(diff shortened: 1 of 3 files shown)


timothyb89:
Also, this no longer builds a static aggregator binary. The workaround from #418 is no longer valid as `alpine:3.7` now includes the affected version of go. The _new_ workaround is to turn off static builds and include librdkafka in the image. I'm sure this has other drawbacks, but it seems to work alright for now. The image should also be quite a bit smaller.

PR #423, opened 2018-04-11
Use docker's `--password-stdin` in CI

timothyb89:
New versions of the Docker CLI warn about using `--password` and
require user confirmation, breaking CI. This switches to using
`--password-stdin`.


timothyb89:
We'll need to include `!push monasca-aggregator` in the merge commit message.

PR #425, opened 2018-04-12
Fix issue with failing travis CI: Integrate change #423 from monasca/monasca-docker

mattibf:
Travis CI jobs for monasca-docker are failing because docker login requires a confirmation.
Integration of pull request #423 provides password in a different way and thus avoids the confirmation  message


mattibf:
Done for the wrong branch - pls. ignore

PR #426, opened 2018-04-26
Make Kafka's `connections.max.idle.ms` configurable

timothyb89:

### DIFF
--- a/kafka/build.yml
+++ b/kafka/build.yml
@@ -1,6 +1,6 @@
 repository: monasca/kafka
 variants:
-  - tag: 0.9.0.1-2.11-1.1.6
+  - tag: 0.9.0.1-2.11-1.1.7
     args:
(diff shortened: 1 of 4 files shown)


kafka/build.yml
```diff
@@ -1,6 +1,11 @@
 repository: monasca/kafka
 variants:
-  - tag: 0.9.0.1-2.11-1.1.6
+  - tag: 0.9.0.1-2.11-1.1.7
     args:
       KAFKA_VERSION: "0.9.0.1"
       SCALA_VERSION: "2.11"
+
+      # it seems kafka 0.9.x has been removed from mirrors...
```
matrixik: Why not adding proper URLs straight to `Dockerfile`? It's not like they will change any time soon.

In devstack we already use archives:
https://github.com/openstack/monasca-api/blob/master/devstack/settings#L172-L173

PR #430, opened 2018-06-07
[keystone-init] add support for assigning domain-level roles to groups

timothyb89:
This adds support for a new `domain_roles` attribute in group
definitions.

### DIFF
--- a/keystone-init/keystone_init.py
+++ b/keystone-init/keystone_init.py
@@ -594,7 +594,7 @@ def grant_group_project_role(client, role_id, group, project):
 
 @retry()
 def grant_group_domain_role(client, role_id, group, domain):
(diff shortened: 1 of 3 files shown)


keystone-init/keystone_init.py
```diff
@@ -554,6 +592,23 @@ def grant_group_role(client, role_id, group, project):
     client.roles.grant(role_id, group=group, project=project)
 
 
+@retry()
+def grant_group_domain_role(client, role_id, group, domain):
+    """Grant a role to a group on a project
```
matrixik: Should probably be domain here, not project.

timothyb89: good catch, fixed

PR #431, opened 2018-06-08
Add support for self-monitoring

zreigz:


zreigz:
PTAL @matrixik @witekest @Brandstetter

PR #433, opened 2018-06-13
Update monasca-watcher golang library

aadams19:
The golang library 1.10.3 is protected against the zip slip vulnerability.
Signed-off-by: Andrea Adams <aadams@hpe.com>


aadams19:
More invistigation is needed because go 1.9.4 is still getting installed
```
(20/26) Installing git (2.15.2-r0)
(21/26) Installing go (1.9.4-r0)
(22/26) Installing glide (0.13.1-r0)
```

matrixik:
Probably similarly like in #434 `glide` is installing `go` as a dependence.

matrixik:
@aadams19 Could you install in similar way that was done in #434 ?

PR #434, opened 2018-06-13
Update build process and go verion

rbrndt:


monasca-aggregator/Dockerfile
```diff
@@ -1,27 +1,47 @@
...
-  apk add --no-cache --virtual build-dep \
-    librdkafka-dev git go glide make g++ openssl-dev && \
-  mkdir -p $GOPATH/src/github.com/monasca/monasca-aggregator && \
+run apk add --no-cache openssl py2-jinja2 && \
+  apk add --no-cache --virtual build-dep git go glide make g++ openssl-dev
+
+# librdkafka is not available in a stable alpine release
+run apk add --update --no-cache \
+        libressl \
+    --repository http://nl.alpinelinux.org/alpine/edge/main  \
```
matrixik: Use one `ARG` for `http://nl.alpinelinux.org/alpine` so it could be changed in one place only.
Also looking at https://mirrors.alpinelinux.org/ I think that `uk.alpinelinux.org` would be faster (but `nl` is also fine with me).

monasca-aggregator/Dockerfile
```diff
@@ -1,27 +1,47 @@
...
 arg AGGREGATOR_BRANCH=master
 
 env GOPATH=/go
 
-run apk add --no-cache openssl py2-jinja2 librdkafka && \
-  apk add --no-cache --virtual build-dep \
-    librdkafka-dev git go glide make g++ openssl-dev && \
-  mkdir -p $GOPATH/src/github.com/monasca/monasca-aggregator && \
+run apk add --no-cache openssl py2-jinja2 && \
+  apk add --no-cache --virtual build-dep git go glide make g++ openssl-dev
```
matrixik: Please, remove installing `go` from here. It's installing `go` in version `1.9.4`.
`(17/25) Installing go (1.9.4-r0)`

monasca-aggregator/Dockerfile
```diff
@@ -1,27 +1,47 @@
...
-  cp monasca-aggregator /monasca-aggregator && \
-  mkdir /specs && \
-  cp aggregation-specifications.yaml /specs && \
-  apk del build-dep && \
-  cd / && \
-  rm -rf /go
+  go build
+
+# librdkafka is not available in a stable alpine release
+from alpine:3.7
```
matrixik: Thank you for making this two step build.

monasca-aggregator/Dockerfile
rbrndt: Interesting, even after removing go from the list, it installs as a dependency of glide

monasca-aggregator/Dockerfile
rbrndt: agreed, will add. I have no preference on which mirror it uses, will switch to uk

rbrndt:
Changed the build to grab glide from github and build/install locally

monasca-aggregator/Dockerfile
matrixik: Well, you replaced http://nl.alpinelinux.org/alpine/edge/main with `ALPINE_EDGE_REPO` in only one place. What about here and line 48?

monasca-aggregator/Dockerfile
```diff
@@ -1,27 +1,54 @@
...
+  apk add --no-cache --virtual build-dep git make g++ openssl-dev
+
+# librdkafka is not available in a stable alpine release
+run apk add --update --no-cache \
+        libressl \
+    --repository $ALPINE_EDGE_REPO  \
+    && apk add --update --no-cache \
+        librdkafka \
+        librdkafka-dev \
+    --repository $ALPINE_EDGE_REPO
```
matrixik: Previously here you used `edge/community` repo. `edge/main` is also OK? Do they have different versions of libraries?

monasca-aggregator/Dockerfile
```diff
@@ -1,27 +1,54 @@
...
+
+run git clone https://github.com/Masterminds/glide.git $GOPATH/src/github.com/Masterminds/glide && \
```
matrixik: I think that `github.com/Masterminds/glide` part also could be extracted to one variable (sorry that sometimes I'm PITA, I just don't like to change stuff in many places if I can change in one).

PR #440, opened 2018-06-26
Add missing dependencies

zreigz:


zreigz:
PTAL @matrixik

PR #442, opened 2018-06-27
Update log-api version

zreigz:


zreigz:
PTAL @adriancz @matrixik

PR #445, opened 2018-07-03
Add dependencies to build monasca-common

zreigz:


zreigz:
PTAL @matrixik @adriancz @witekest

Brandstetter:
tested for monasca-base and log-api image -->OK

PR #480, opened 2018-12-07
Fix multiline logs

martinchacon:
Added multiline-log feature

- From v3.2.5 logspout supports multiline-logs [1]

[1] https://github.com/gliderlabs/logspout/tree/v3.2.5#multiline-logging

### DIFF
--- a/logspout/README.md
+++ b/logspout/README.md
@@ -30,10 +30,10 @@ to the Monasca Log Agent.
 Configuration
 -------------
 
-|        Variable         |             Default             |             Description             |
-|-------------------------|---------------------------------|-------------------------------------|
-| `MONASCA_LOG_AGENT_URI` | `log-agent:5610`                | Monasca Log Agent connection string |
-| `ROUTE_URIS`            | `logstash+tcp://log-agent:5610` | Logstash connection string          |
+|        Variable         |                 Default                   |             Description             |
+|-------------------------|-------------------------------------------|-------------------------------------|
+| `MONASCA_LOG_AGENT_URI` | `log-agent:5610`                          | Monasca Log Agent connection string |
+| `ROUTE_URIS`            | `multiline+logstash+tcp://log-agent:5610` | Logstash connection string          |
 
 Additional files
 ----------------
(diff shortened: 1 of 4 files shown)


logspout/modules.go
```diff
@@ -6,4 +6,5 @@ import (
 	_ "github.com/gliderlabs/logspout/transports/tcp"
 	_ "github.com/gliderlabs/logspout/transports/udp"
 	_ "github.com/looplab/logspout-logstash"
+        _ "github.com/gliderlabs/logspout/adapters/multiline"
```
matrixik: File should be formated with gofmt. You could also do it online on https://play.golang.org/

logspout/build.yml
```diff
@@ -1,7 +1,7 @@
 repository: monasca/logspout
 variants:
-  - tag: 0.0.2
+  - tag: 0.0.3
```
matrixik: Please, bump it to `0.1.0`.

logspout/README.md
```diff
@@ -13,6 +13,7 @@ Tags
 Images in this repository are tagged as follows:
 
 * `0.0.1`: standard [SemVer][3].
+* `0.0.3`: added multiline-logs support.
```
matrixik: I don't think it's needed here, this place is more like description of used tags what are they for.

martinchacon: Done

logspout/build.yml
martinchacon: Done

logspout/modules.go
martinchacon: Done

.env
```diff
@@ -34,6 +34,6 @@ MON_ELASTICSEARCH_CURATOR_VERSION=4.3.1-1.0.0
 MON_LOG_API_VERSION=master-20180627-055222
 MON_KIBANA_VERSION=4.6.3-master
 MON_LOG_AGENT_VERSION=0.0.5
-MON_LOGSPOUT_VERSION=0.0.2
+MON_LOGSPOUT_VERSION=0.1.0
```
matrixik: This change needs to be in different PR.

PR #482, opened 2018-12-18
Remove mysql from http_check

matrixik:
We already have separate MySQL check and http_check
for same service is not working.


matrixik:
Needs `!push monasca-agent-collector`

Issue #485, opened 2018-12-25
aggregator lack of log level config property


zhangjianweibj:
aggregator can config log level in the config.yaml,but monasca-docker project lack of this property.this property should be available to users.
![image](https://user-images.githubusercontent.com/18585129/50418506-ad60cf00-0866-11e9-958e-17c219d1ae46.png)

![image](https://user-images.githubusercontent.com/18585129/50418514-bce01800-0866-11e9-96d3-bf3b263351a0.png)

matrixik:
Thank you for report and fix.

PR #486, opened 2018-12-25
aggregator lack of loglevel property

zhangjianweibj:
https://github.com/monasca/monasca-docker/issues/485

hello,aggregator can config log level in the config.yaml,but monasca-docker project lack of this property.this property should be available to users.


matrixik:
I make it more consistent with rest of monasca-docker repository. So use `LOG_LEVEL` and bump image version.

PR #499, opened 2019-02-15
Update monasca-api to master and fix build

matrixik:
Include missing dependencies. `pykafka` was removed from upper
constraints file.

Old code used for building monasca-api
https://review.openstack.org/#/c/417163 is not building anymore because
depended library included breaking changes.

Use newer version of monasca/python image.


matrixik:
Needs `!push monasca-api-python`

PR #512, opened 2019-04-11
Docker support statsd disbaled

zhangjianweibj:
monasca-notification support disabled option for statsd.

i hava resolved in 'https://review.openstack.org/#/c/649969/'


matrixik:
Looks like this image here is still building from old change https://review.openstack.org/#/c/461917/ and building stopped worked correctly. I will need some more time to investigate.

matrixik:
OK, master branch from https://github.com/openstack/monasca-notification is now pushed to https://hub.docker.com/r/monasca/notification/tags/ . It will be pushed every 24h to always have up to date image running with Python 3. 

With the image that is here looks like there was quite a few changes to dependencies and it's a problem to fix it. I'm not sure if I will be able to do it today and from tomorrow I have 2 weeks of vacations.

zhangjianweibj:
hava a nice vacation

matrixik:
Sorry, I won't be able to fix it here. There are too many problems... At last `master` now is new on docker hub from most monasca repos.
https://hub.docker.com/u/monasca

matrixik:
Please, use images build straight from openstack repositories.

PR #515, opened 2019-04-18
Remove old Docker tags from build scripts

matrixik:
Old tags removed from:
- monasca-api-python
- monasca-log-api
- monasca-notification
- monasca-persister-python

This old tags are still available on Docker Hub, now this images are
build in Zuul every 24 h and available with `master` tag so remove
`master` tag from tags used in monasca/monasca-docker to avoid
collisions.

### DIFF
--- a/monasca-api-python/build.yml
+++ b/monasca-api-python/build.yml
@@ -1,8 +1,6 @@
 repository: monasca/api
 variants:
-  - tag: master
-    aliases:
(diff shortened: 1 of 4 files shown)


matrixik:
Shit, this images build but are not working any more because of changes in dependencies...

Sooo, I don't need to care so much about colliding Docker tags because new changes won't pass tests...

PR #519, opened 2019-05-31
So far use STORM_UI_HOST isn't suitable for docker, because container…

like-inspur:
… ip will be changed. So it should expose address using 0.0.0.0

### DIFF
--- a/storm/templates/storm.yaml.j2
+++ b/storm/templates/storm.yaml.j2
@@ -43,7 +43,7 @@ nimbus.seeds:
 nimbus.thrift.port: {{ NIMBUS_THRIFT_PORT | default('6627') }}
 nimbus.childopts: "{{ NIMBUS_CHILDOPTS }}"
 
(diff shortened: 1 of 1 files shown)


storm/templates/storm.yaml.j2
```diff
@@ -43,7 +43,7 @@ nimbus.seeds:
 nimbus.thrift.port: {{ NIMBUS_THRIFT_PORT | default('6627') }}
 nimbus.childopts: "{{ NIMBUS_CHILDOPTS }}"
 
-ui.host: "{{ STORM_UI_HOST | default ('127.0.0.1') }}"
+ui.host: 0.0.0.0
```
matrixik: I would prefer to allow operator to change `ui.host` if he want to, so instead of removing all configuration option just replace default value: `"{{ STORM_UI_HOST | default ('0.0.0.0') }}"`

Issue #521, opened 2019-07-27
mysql init error


zhangjianweibj:
MySQL [mon]> grant all on mon.* to notification@'%' identified by 'password';
ERROR 1133 (42000): Can't find any matching row in the user table

reference:

https://stackoverflow.com/questions/17597144/cant-find-any-matching-row-in-the-user-table

PR #522, opened 2019-07-27
Fix mysql init

zhangjianweibj:
mysql init bug fix:

https://github.com/monasca/monasca-docker/issues/521


zhangjianweibj:
ok.very thanks.

PR #525, opened 2019-08-15
Fix thresh config bug

zhangjianweibj:
generate thresh-config.yml file.


matrixik:
Thank you, we also have almost ready image in https://review.opendev.org/#/c/649298

